PR #37, opened 2014-12-31
Add site news and associated atom feed

WasabiFan:
This PR adds a "news" page that displays a feed of posts (part of the discussion in ev3dev/ev3dev#109). There is also an atom feed of the same posts. I included a "lorem ipsum" news item to test with but we can (and probably should) replace it with an actual post before merging the PR.

To add this, there were a few larger things that I changed:
- I added a "community" drop-down that includes the news, project gallery, and contributing menu items. If I had added another without collapsing these, the nav would have been much too wide for smaller screens.
- I deleted the old `_posts` folder and moved the projects to a new `_posts` folder inside the existing projects folder. This both eliminates the need for explicit category definition (because Jekyll derives the set of categories by the directory structure above the posts) and allows us to add multiple post categories and keep them separate.

Neither of these changes should break anything technically, although we will need to update the "adding a project" guide with the new path.

My `gh-pages` branch has been updated with these changes as well.


dlech:
I think this looks super!

Just one more nit-pick. I think "Issues" should be moved to the bottom of the "Community" menu.

WasabiFan:
I just made those last few adjustments and added text to the initial post.

dlech:
This just occurred to me, but it would be nice to have news separated out in the quick nav results too.

dlech:
Right now, in the quick nav results it shows the title and the author, but it would be better to show the date instead of the author.

WasabiFan:
I was planning to fix that on a later date (in a separate PR) but it shouldn't take more than 10 minutes. Expect another commit soon.

WasabiFan:
I added a 'news' section to the quick nav dropdown, and also added the date below each result.

dlech:
Once again, THANKS VERY MUCH!

rhempel:
Yes, thanks very much! It makes the site that much better for everyone.

PR #38, opened 2015-01-22
stated that the Ev3 should be connected

moriarty:
The Ev3 should be connected to the host computer at some point before step 4, so I added a reminder in step 3.


dlech:
Good idea. The `v` in `EV3` should be capitalized though.

moriarty:
@dlech, I just fixed the capital V via the browser editor. If you prefer I squash the commits together let me know.

dlech:
For a code repository, I  would say yes to squashing, but for the website, I don't really care.

dlech:
Thanks!

PR #39, opened 2015-01-22
Create setting-up-ethernet-over-usb-on-linux-command-line.md

moriarty:
From ev3dev/ev3dev#253 

This is the same as the NetworkManager file with 3 steps replaced with 1 step.

I can alternatively add optional steps to the Linux with Network Manager page.

### DIFF
--- a/docs/tutorials/setting-up-ethernet-over-usb-on-linux-command-line.md
+++ b/docs/tutorials/setting-up-ethernet-over-usb-on-linux-command-line.md
@@ -1,0 +1,82 @@
+---
+title: Setting Up Ethernet Over USB on Linux with Command Line
+subject: Ethernet Over USB
+---
+
+1.  On the EV3, first verify that the CDC driver is enabled. In brickman,
+    open the *USB* settings and make sure *CDC* is selected and active.
+
+    {% include screenshot.html source="/images/brickman/usb-cdc-active.png" %}
+
+2.  Then, in *Networking*, make sure *Gadget* is enabled.
+
+    {% include screenshot.html source="/images/brickman/networking-technologies-gadget-selected.png" %}
+
+3.  On your host computer, with the Ev3 connected via USB, open a *Terminal* and run
+
+        user@host:~$ ifconfig | grep usb
+  
+    Assuming there is only ```usb0``` the output should look like:
+  
+        usb0      Link encap:Ethernet  HWaddr 12:16:53:46:2e:42 
+  
+    Now set the ip address of ```usb0``` to ```10.42.0.1``` with the following command. The ```/24``` lets the host computer know the range of ip addresses connected to this interface.
+  
+        user@host:~$ sudo ip ad add 10.42.0.1/24 dev usb0
+
+4.  Now, we need to assign an IP address to our EV3. In brickman, go to
+    *Networking* and select *Manage connections...*, then select the *Wired*
+    connection.
+
+    {% include screenshot.html source="/images/brickman/networking-connections-wired-only.png" %}
+
+    On the *IPv4* tab, select *Change...*.
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/setting-up-ethernet-over-usb-on-linux-command-line.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Setting Up Ethernet Over USB on Linux with Command Line
```
dlech: What tools are you using? The name of the tools should be added to the title. For example, it is possible to configure NetworkManager from the command line. If I was doing this, I would make the title something like "Setting Up Ethernet Over USB on Linux using NetworkManager via the Command Line"

docs/tutorials/setting-up-ethernet-over-usb-on-linux-command-line.md
```diff
@@ -0,0 +1,82 @@
...
+
+2.  Then, in *Networking*, make sure *Gadget* is enabled.
+
+    {% include screenshot.html source="/images/brickman/networking-technologies-gadget-selected.png" %}
+
+3.  On your host computer, with the Ev3 connected via USB, open a *Terminal* and run
+
+        user@host:~$ ifconfig | grep usb
+  
+    Assuming there is only ```usb0``` the output should look like:
```
dlech: Code snippets actually only need one backtick like this: ``` `usb0` ``` (unless the snippet contains a backtick as well like in this comment).

docs/tutorials/setting-up-ethernet-over-usb-on-linux-command-line.md
```diff
@@ -0,0 +1,82 @@
...
+  
+        usb0      Link encap:Ethernet  HWaddr 12:16:53:46:2e:42 
+  
+    Now set the ip address of ```usb0``` to ```10.42.0.1``` with the following command. The ```/24``` lets the host computer know the range of ip addresses connected to this interface.
+  
+        user@host:~$ sudo ip ad add 10.42.0.1/24 dev usb0
```
dlech: I am not familiar with setting up the network this way. Does this persist across reboots? Does it make changes to `/etc/network/interfaces`?

dlech:
> I can alternatively add optional steps to the Linux with Network Manager page.

No, I would rather have a separate page like this.

docs/tutorials/setting-up-ethernet-over-usb-on-linux-command-line.md
moriarty: No, actually it's not even persistent when the USB cable is removed and re inserted. 
It doesn't change `/etc/network/interfaces`.

I will find a way to do it persistently - edit /etc/network/interfaces is one, I'll check if there is another.

docs/tutorials/setting-up-ethernet-over-usb-on-linux-command-line.md
moriarty: Thank you! I've done so many code snippets in commit messages and comments and have always been using three.

docs/tutorials/setting-up-ethernet-over-usb-on-linux-command-line.md
moriarty: Okay, The RaspberryPi doesn't have network-manager installed by default. 
When I find a persistent method then I will update the title accordingly.

moriarty:
Sorry I forgot about this completely.

dlech:
So do you want to finish it or should we just close the pull request?

moriarty:
I don't know when I'll be able to finish it :slightly_frowning_face: Sorry.

PR #40, opened 2015-01-23
Fix typo in getting-started.md

WasabiFan:


dlech:
I really wish spell check would work when doing in-browser edits. :disappointed:

PR #41, opened 2015-01-24
Users should be in input group

ddemidov:
In order to have access to brick buttons, users should be in `input` group.


dlech:
:+1:  We don't need to worry about key loggers on the EV3 brick, I don't think.

PR #44, opened 2015-02-21
Add Clojure to the list of extra languages.

annapawlicka:
I've created a Clojure wrapper around ev3dev API. It uses ssh to send commands (as starting Clojure REPL on ev3 is not possible due to its heavy system requirements). It's still WIP but it's got most of the functionality covered.


dlech:
Looks like you are off to a good start. Do you think we should mention here that it is "remote control only" since it does actually run on the EV3?

annapawlicka:
Yes, good idea. Thx!

I'm trying to run [pixie](https://github.com/pixie-lang/pixie) on ev3dev as I write this - a potential light-weight alternative that's inspired by Clojure and runs on ARM. But Clojure itself is remote control only I'm afraid.

dlech:
There are so many possibilities with ev3dev, the fun may never end!

PR #46, opened 2015-03-06
Couple of fixes

MolarAmbiguity:


dlech:
The troubleshooting section does not seem that helpful since it does not tell you how to fix the problem. Could you add more? What do I do if ping fails? What do I do if ping succeeds?

dlech:
I cherrypicked the fix for Clojure. I have made some changes to the networking tutorials, so I can't merge the troubleshooting section. I'm thinking it might be better for troubleshooting to have it's own page anyway as in `/docs/troubleshooting/ssh` and on the `/docs` page, add a menu item that says "Basic Troubleshooting".

MolarAmbiguity:
Ok that sounds good.

PR #47, opened 2015-03-11
Added Tutorial for Removing SD card

MolarAmbiguity:
p.s you can ignore the troubleshooting bit, it is still under development.


docs/tutorials/sd-card-removal.md
```diff
@@ -0,0 +1,25 @@
+---
+title: SD card Removal
```
dlech: When I first read this, it made me think it was about how to "safely remove" (via software) the SD card. Might be better to say this is about physically removing the card.

docs/tutorials/sd-card-removal.md
```diff
@@ -0,0 +1,25 @@
...
+
+You've probably put your SD card into the EV3 by now and successfully booted EV3dev. Now you want to pull the card out and check that the default firmware is still working. You push on the card, expecting it pop out but it stubbornly remains in your robot. There are a few ways to remove the card from the EV3, they are as follows.
+
+Method 1: Tape
+
+Have a small bit of tape connected to the sd card. You can pull on this to easily remove the card.
+
+{% include screenshot.html source="/images/TapeSD.png" %}
+
+This is probably the best method for removal of the card.
```
dlech: Should probably say "gently" - especially when using pliers.

docs/tutorials/sd-card-removal.md
```diff
@@ -0,0 +1,25 @@
...
+
+Method 2: Pliers
+
+Grab the card between some small pliers/tweezers.
+
+
+Method 3: Disassembly
```
dlech: This is not practical.

dlech:
Thinking about this a bit more, I think I would rather like to see only method 1 as a "tip" on the getting started page. Or actually we could make it an actual step by itself and include the piece of tape on the list of things you needed. What do you think?

The photograph is very good. Could you take a photo of the SD card by itself too so we can see how the tape is attached?

docs/tutorials/sd-card-removal.md
MolarAmbiguity: I know, I was trying to provide a complete list of ways to remove the card.

MolarAmbiguity:
Thank-you for your feedback

dlech:
Everything here is ev3dev-specific. Also, using tape to aid in removal of the SD card is very good advice. I would like as many people as possible to know about it, which is why I suggested putting it on the getting started page.

MolarAmbiguity:
What I meant by "general EV3 user documentation" was a separate wiki or website for general EV3 stuff. Maybe I should post something here: http://bricks.stackexchange.com ?

dlech:
The most active "general EV3" site that I know of is [this facebook group](https://www.facebook.com/groups/legomindstorms/), but it doesn't really work well for that kind of documentation. stackexchange is in a question/answer format, so that doesn't really seem appropriate either. It seems that this kind of information is mostly found on blogs [like this](http://botbench.com/blog/2013/10/29/ev3-adding-a-pull-tab-to-your-micro-sd-card/).

MolarAmbiguity:
Its ok if you refuse this pull, I think I need to work on it some more. (And also figure out exactly how git works)

dlech:
I am not going to merge it as-is. If you would like to keep working on it more, I think the bit about using tape on the SD card has potential. Or if you want to start over or write it off as a loss, you can close the pull request.

MolarAmbiguity:
Ok, I'm going to rework this later on. I didn't see the button to close pull request before.

PR #50, opened 2015-03-24
Added informations about the magic number in the battery icon

theZiz:
I thought somewhere in the wiki should be explained, what the "8.0" in the picture means. I first thought this is some kind of version and my fresh installed Debian is already oudated (becase I had only "Version" 7.39). Even if you know, that this is the remaining energy, it isn't obvious, that the brick will die at ~5V.


dlech:
We really need a full explanation of brickman somewhere, but this is good for now. Thanks.

PR #52, opened 2015-04-11
 Use tile structure instead of bullets on projects page  

WasabiFan:
Live here: http://wasabifan.github.io/ev3dev.github.io/projects/


dlech:
Wow. That's much better.

Can we get a little more space between the image and the text?

I'm also thinking it might be good to add the `excerpt:` line in the frontmatter to the sample project with an explanation of when it might bee needed.

dlech:
And I broke this by merging your other pull request.

WasabiFan:
I just pushed those modifications.

PR #54, opened 2015-04-11
Constrain logo

WasabiFan:
The new org logo is bigger than the old one, so it overlapped the rest of the header.


dlech:
I just pushed some changes that took care of this. Logo is now the same size always.

PR #59, opened 2015-05-12
Update getting-started.md

theZiz:
Added a C library

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -207,9 +207,7 @@ Here are some guides for using each of the major components.
 There are a number of programming languages available to use. The brick can
 run almost all languages that any other Linux distro can, so your favorite
 language is probably supported. Language bindings have already been written
-for some languages:
-
-* [Libraries](/docs/libraries)
+for many languages. You can learn more about the available libraries [here](/docs/libraries).
 
 If the language you want isn't listed, you still can use it, but you'll have to
 do more of the heavy lifting yourself using the guides above. Once you get the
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -215,6 +215,7 @@ for some languages:
 - [Google Go](https://github.com/ldmberman/GoEV3)
 - [Node.js](https://github.com/WasabiFan/ev3dev-NodeJS)
 - [Python](https://github.com/topikachu/python-ev3)
+- [C](https://github.com/theZiz/ev3c)
```
dlech: Can you please put this in alphabetical order?

dlech: On second though, maybe we should replace this list with a link to the other (libraries) page so that we don't have to maintain 2 lists of the same information.

WasabiFan: I think it should be replaced with a link as well -- this list is pretty outdated.

theZiz:
How about this?

WasabiFan:
I think that the link should be surrounded by a sentence and put in the paragraph above; something like "you can learn more about the available libraries [here]". Also, I'm worried about the relative path to the docs page... Have you tested this being rendered by GitHub?

theZiz:
I just did the same like e.g. the Linux command line link. In the preview, it doesn't seem to work, but in the final page it does and should also for my link.
About the text, I am not good at writing English text, so I will just copy your suggestion. If you want another formulation you need to write it yourself. :)

dlech:
Good enough for me.

dlech:
For future reference though, the "right way" to do links is `{{ github.site.url }}/docs/libraries`, then jekyll will fill in the proper prefix so that the links work when testing in your own fork on GitHub.

PR #61, opened 2015-05-18
Update official bindings list

ddemidov:
Removed R, added Python, moved vala to `Extra languages` section. R should probably be mentioned in `Extra languages` as well, but it only exists in a branch of the official repository, so not sure how to link to it.

### DIFF
--- a/docs/libraries.md
+++ b/docs/libraries.md
@@ -32,6 +32,6 @@ We also have many great contrubutors that are maintaining extra libraries for la
     * [Google Go](https://github.com/ldmberman/GoEV3) updated for ev3dev-jessie by @ldmberman, [original](https://github.com/mattrajca/GoEV3) by @mattrajca
     * [Python](https://github.com/topikachu/python-ev3) by @topikachu
     * [C (with optional Perl, Python and Ruby bindings)](https://github.com/in4lio/ev3dev-c) by @in4lio
-    * [C](https://github.com/theZiz/ev3c) by @Ziz
+    * [C](https://github.com/theZiz/ev3c) by @theZiz
     * [Clojure](https://github.com/annapawlicka/clj-ev3dev) by @annapawlicka
     * [vala](https://github.com/dlech/ev3dev-lang-glib) by @dlech and @WasabiFan
--- a/docs/libraries.md
+++ b/docs/libraries.md
@@ -42,7 +42,9 @@ for ev3dev is written using this library.
 
 
 ## Extra languages
-We also have many great contrubutors that are maintaining extra libraries for languages not included in our other repository. 
+We also have many great contrubutors that are maintaining extra libraries for
+languages not included in our other repository.  Note that some of these
+libraries may be outdated due to the fast development cycle of ev3dev.
 
 * Extra languages:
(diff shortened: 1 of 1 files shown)


WasabiFan:
I don't have a problem with it, I was just pointing out that I really haven't been much help to @dlech with that binding.

WasabiFan:
I don't have a problem with it, I was just pointing out that I really haven't been much help to @dlech with that binding.

ddemidov:
From the github history it looked like you started the work :). I switched your names. Would that be ok?

WasabiFan:
Yeah, it's fine ;) I didn't intend to make you change anything.

docs/libraries.md
```diff
@@ -34,3 +34,4 @@ We also have many great contrubutors that are maintaining extra libraries for la
     * [C (with optional Perl, Python and Ruby bindings)](https://github.com/in4lio/ev3dev-c) by @in4lio
     * [C](https://github.com/theZiz/ev3c) by @Ziz
     * [Clojure](https://github.com/annapawlicka/clj-ev3dev) by @annapawlicka
+    * [vala](https://github.com/dlech/ev3dev-lang-glib) by @dlech and @WasabiFan
```
dlech: For vala, which actually works with any language that supports gobject introspection, the code has been merged into https://github.com/ev3dev/ev3devKit

dlech:
> R should probably be mentioned in Extra languages as well

There are several other links that have not been keeping up with the kernel changes. I think we could make a new "broken" or "out of date" section and add those there. I know that @in4lio and python-ev3 have been keeping up and @Ziz is new, but I am not sure about the others.

ziz:
I think you want it to be "theZiz", not "Ziz", since I'm fairly sure I have not contributed anything to this project. :)

dlech:
Oops, wrong ziz. :blush: 

I guess we better fix that in line 35 as well.

theZiz:
Oh, sorry, I didn't knew, that this makes a github link!

ddemidov:
> There are several other links that have not been keeping up with the kernel changes. I think we could make a new "broken" or "out of date" section and add those there. 

Tracking status of third-party libraries sounds like a lot of work to me. How about adding a disclaimer like ``Some of those libraries may be outdated due to the fast development cycle of ev3dev``?

ddemidov:
> I guess we better fix that in line 35 as well.

Fixed in 7493f98

docs/libraries.md
ddemidov: Since the toolkit is maintained by you, and is used for brickman development, I feel it deserves a separate section. It definitely does not belong to "Extra languages". So I've added the "Programming toolkit for ev3dev" in ab76f26. What do you think about it?

dlech:
> Tracking status of third-party libraries sounds like a lot of work to me. How about adding a disclaimer like Some of those libraries may be outdated due to the fast development cycle of ev3dev?

I think that is a good idea. Let's do that instead.

ddemidov:
> Let's do that instead.

Done in 17ba55a

dlech:
Thanks!

PR #64, opened 2015-06-02
Add 'Projects' item under 'Community' menu

ddemidov:
Also replace last project with sample project on 'Share' page.

This was proposed by @cavenel on the irc channel, I think this is a great idea too.

### DIFF
--- a/share.md
+++ b/share.md
@@ -13,8 +13,8 @@ ev3dev.org Projects Page
 We currently have a [projects page] where you can browse projects that have been
 built using ev3dev. You can add your own too!
 
-{% assign project = site.categories.projects | first %}
-Here is the most recent project ([more about this project...]({{ site.github.url }}{{ project.url}})).
+{% assign project = site.categories.projects | last %}
+Here is the sample project ([more about this project...]({{ site.github.url }}{{ project.url}})).
 
 ###{{ project.title }}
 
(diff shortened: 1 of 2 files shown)


share.md
```diff
@@ -13,8 +13,8 @@ ev3dev.org Projects Page
 We currently have a [projects page] where you can browse projects that have been
 built using ev3dev. You can add your own too!
 
-{% assign project = site.categories.projects | first %}
-Here is the most recent project ([more about this project...]({{ site.github.url }}{{ project.url}})).
+{% assign project = site.categories.projects | last %}
+Here is the sample project ([more about this project...]({{ site.github.url }}{{ project.url}})).
```
dlech: I think we should also have a [link](https://raw.githubusercontent.com/ev3dev/ev3dev.github.io/master/projects/_posts/2014-03-21-Example-Project.md) to the "source" of the sample project on GitHub so that it can easily be copied.

ddemidov: Do you mean instead of the project subpage (`{{ project.title }}`/`{{ project.subpage }}`)?

ddemidov: Ah, I overlooked 'also'.

dlech: I mean in addition to. Somewhere we have instructions on how to create your own project page. Perhaps we could just include it here instead.

ddemidov: I'll try to look at this (and #65) in a couple of hours.

ddemidov:
Contained in #65

PR #65, opened 2015-06-02
Do not include example project into projects page

ddemidov:
The link to the project is still present on the Share page.

This is a result of irc discussion with @cavenel.

### DIFF
--- a/projects/index.md
+++ b/projects/index.md
@@ -12,23 +12,25 @@ Would you like your project added to the list? We would!
 <ul id="projects-list">
     {% for post in site.posts %}
         {% if post.categories contains "projects" %}
-            <li>
-                <div class="project-tile-title">
-                    <h4>
-                        <a href="{{ site.github.url }}{{ post.url }}">{{ post.title }}</a>
-                    </h4>
-                </div>
-                {% assign num_words = 40 %}
-                {% if post.youtube_video_id %}
-                    {% assign num_words = 20 %}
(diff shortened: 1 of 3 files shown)


projects/index.md
```diff
@@ -12,23 +12,25 @@ Would you like your project added to the list? We would!
...
-                    {% assign num_words = 20 %}
-                    <a href="{{ site.github.url }}{{ post.url }}">
-                        <img class="project-tile-img" src="http://img.youtube.com/vi/{{post.youtube_video_id}}/mqdefault.jpg" />
-                    </a>
-                {% endif %}
-                <span class="project-tile-excerpt">
-                    {{ post.excerpt | truncatewords: num_words }}
-                </span>
-            </li>
+            {% unless post.title contains "Example" %}
```
dlech: What if someone else uses the word "Example" in their title? Can we make this a more exact match?

ddemidov: I am not very familiar with liquid templates syntax. I tried to use `post.title != "Example Project"`, but that did not work for some reason. Do you have any suggestions?

dlech: Looking at <http://jekyllrb.com/docs/variables/#page-variables>, there is a `page.id` variable that provides a unique identifier for the page. You can add `{{ page.id }}` to the page temporarily to see what the actual id is.

I don't see any reason why `==` or `!=` should not work.

ddemidov:
060b1f4 uses post.id to exclude the example project.

PR #66, opened 2015-06-02
'Adding a new project' tutorial

ddemidov:
This creates `Adding a new project` tutorial and links to it from the 'Share' page. This is converted from the old [wiki page](https://github.com/ev3dev/ev3dev/wiki/Posting-A-Project) (with updated images and links).


ddemidov:
Do not merge it yet; the link does not work for some reason.

docs/tutorials/adding-new-project.md
```diff
@@ -0,0 +1,158 @@
+Adding a new project
+------------------------
+
+We currently have a [projects page][] where you can browse projects that have
```
dlech: the trailing `[]` on the link is not needed.

ddemidov:
Ok, should be fixed in 3a254ad.

docs/tutorials/adding-new-project.md
ddemidov: Fixed in a29546e

docs/tutorials/adding-new-project.md
```diff
@@ -0,0 +1,160 @@
+---
+title: Adding a new project
+subject: Projects
```
dlech: The subject is used to group pages into categories. (the word "categories" is already used for a special purpose by jekyll).

Since most of the categories are "things you can do with your EV3", I'm thinking we should make a new subject for "things you can do with the www.ev3dev.org webiste". I think a subject of "Projects" is too narrow and will only have this one thing listed under it.

So, `subject: "www.ev3dev.org Website"` or `subject: "This Website"`. What do you think?

docs/tutorials/adding-new-project.md
```diff
@@ -0,0 +1,160 @@
...
+well as provide videos, pictures, build instructions, code, and any other media
+that pertains to the project. This is where we explain how you can submit your
+project page to be hosted on the site.
+
+##Overview
+
+All the projects on the website are stored in the [projects folder].
+Each post is saved as a Markdown file, and is automagically converted to HTML
+when you publish your changes (read more about Markdown
+[here][markdown basics]). To submit a page,
```
dlech: The website actually uses the Kramdown markdown engine rather that GitHub flavored markdown. Here, I think we could say "Kramdown" instead of "Markdown" and use this link: <http://kramdown.gettalong.org/quickref.html>.

docs/tutorials/adding-new-project.md
```diff
@@ -0,0 +1,160 @@
...
+willing to add a bit of extra complexity, you have to create a new branch
+called `gh-pages`. Adding this branch will let you preview the whole site, so
+you can see how your page will look when it is published (which is different
+from the markdown preview that you will see on GitHub). To do this, click the
+box where it says "branch: master". It should expand to show you a text box.
+Type the text `gh-pages` in to the box that appears (make sure that it is
+exactly that text; this is important). It should give you the option to create
+a new branch; click that button.
+
+<img src="/images/Website/Branching-on-GitHub.png" style="max-width: 50%;" />
```
dlech: We have a include template for screenshots.

    {% include screenshot.html source="/images/Website/Branching-on-GitHub.png" scale="50" %}

docs/tutorials/adding-new-project.md
```diff
@@ -0,0 +1,160 @@
...
+contents of the file itself.
+
+This file that you have opened is not an actual project. It is a template file
+that is meant to be copied to make new project creation easier. The part we
+want is everything between the "---"s. This is called "frontmatter," and it
+describes everything about a project, such as the author of the project. Select
+and copy the frontmatter out of the template file (make sure that you include
+the dashes). This is what you should have:
+
+    ---
```
dlech: Idea: we can embed GitHub Gists on the website. What if we host the sample page as a Gist so that the content below is always up-to-date and can be copied from here instead of having to view the raw file somewhere else.

https://gist.github.com/benbalter/5555251

ddemidov: The content below is not a full content of the template file. Only the lines between `---` markers are included (and should be copied according to the tutorial). If we move the sample project to a gist, then we would not be able to show it [rendered](http://www.ev3dev.org/projects/2014/03/21/Example-Project/). And if we keep it here, then again we have a possibility of diverging versions.

May be replace the block below with a link to raw md file?

dlech: Good point. Yes, let's replace it with a link to the raw file.

ddemidov: Also the embedded gist looks awful: http://ddemidov.github.io/ev3dev.github.io/docs/tutorials/adding-new-project/. Probably needs to be saved as text file.

ddemidov: Yes, that (kind of) helps. But still, I don't like the idea of embedded gist in this case.

ddemidov: > Yes, let's replace it with a link to the raw file.

Done

dlech:
I think we will call this good enough for now. Thanks for your work.

PR #67, opened 2015-06-24
Create 2015-05-31-Noflo-EdgeFollowing.md

ensonic:


dlech:
Nice. The visual programming reminds me of the Microsoft Robotics Studio. Is the "code" we see on the monitor running in a separate IDE or is it more like a remote webpage served from the EV3?

Also, I fixed up a few things. There was a link that did not work, which I fixed by surrounding it with `<>` (the website uses a different markdown engine compared to GithHub issues and wikis). And I removed a duplicate video. Having the youtube id in the frontmatter causes the video to be displayed before the rest of the content and you had the same video embedded again later.

dlech:
Oh, and you can add this to the list of programming languages at <http://www.ev3dev.org/docs/libraries/> if you like.

ensonic:
The IDE is a pure javascript app (noflo-ui), you can try that at http://app.flowhub.io. The noflo-nodejs app on the ev3 implements the FBP (flowbased programming) protocol (http://noflojs.org/documentation/protocol/). This is what the noflo-ui uses to talk tot he runtime.

What I personally like a lot about this approach is that one actually edits the graph (the programm) live on the target. There is no

rhempel:
Holy cow! So if I understand this correctly, the dataflow diagrams are edited in your browser, and the node.js app/server is running on the EV3? That's pretty neat! 
Kind of reminds me of node editing in Blender.
Nice work! Send me your mailing address and you get an ev3dev.org tile!

ensonic:
@dlech Before I'll add this to the library page. I'll bump my work to run on the new image and I'll port my noflo-legoev3 to the nodejs bindings that you have by now .

@rhempel yep, what you edit in the browser is a view on the the live program. You can click each wire and get a live preview of the data going though it. And to emphasize the js app just loads from e.g. app.flowhub.io, but it is a pure js app, that don't needs internet once you have it running (it is also in the chrome webstore). I think this could be potentially nice for edu together with chromebooks.

PR #69, opened 2015-06-30
Clean up and format CSS

WasabiFan:
This moves the CSS into 5 or 6 separate files and formats all of it. In doing this, I found a few instances where I had changed a class name in one location and not in another (such as the sidebar content wrapper) and fixing those made a few mobile views work much better (it's amazing what happens when all your code is being loaded and evaluated!).

I also moved the CSS link and JS script tags into Jekyll templates to decrease copied code in the header.

There will probably be another wave of fixes coming soon, but I decided that this was a good time to open a PR to make sure it looks good.


WasabiFan:
Just realized that I forgot to uncomment some HTML. I'll fix that and amend the PR as soon as I can -- but that may be in an hour or so.

WasabiFan:
I just fixed that OL that I hadn't uncommented and added some small tweaks. Should be OK now.

dlech:
I didn't open this in an editor to see exactly what you did, but it looks like you cleaned up some whitespace. This is fine, just curious what exactly you did (spaces/tabs).

WasabiFan:
> Why use include files with only one line? It is more obvious to me what is going on without using an include here. 

I did that because I figured that even though it only brings minimal code savings, it means that changing it in one place changes it in each link -- things like the folder path where we store JS might change, and we need to include the root from the GH config every time. Admittedly, it didn't save as much code as I had hoped... but it's marginally better in my opinion.

> I didn't open this in an editor to see exactly what you did, but it looks like you cleaned up some whitespace. This is fine, just curious what exactly you did (spaces/tabs).

I normalized all the formatting and indentation: some of it was formatted with two spaces, some with four, and some with tabs (which are variable). So, I just normalized it to 4 spaces per unit indent.

> Why ordered list ( ol ) and not unordered list ( ul )?

Because the `o` in `ol` was closer to my index finger :wink:? It didn't seem to matter, so I just didn't take the time to think about reasons to use one or the other. As far as I know, the only difference is the default icon (#s vs. dots) and we disable the icon anyway. I guess the header is expected to be in a specific order, right? If there's a reason that I should do it the other way, I'd be happy to change it -- I just wanted to wrap the rampant `li`s in a list element.

dlech:
> things like the folder path where we store JS might change

I find that highly unlikely. I'd rather go for better readability here and omit d6fbbb6.


> I just normalized it to 4 spaces per unit indent.

Excellent.


> I'd be happy to change it

Please. In my experience, it seems to be standard to use `ul` for menus rather than `ol`.

WasabiFan:
@dlech Just made those minor changes -- can you take another look?

P.S. The CC logo is returning a `301 (Moved Permanently)` right now. I included a commit to change it to the URL that the response headers were advising me of.

dlech:
All looks good. Thanks!

PR #75, opened 2015-11-21
Calling all robots

WasabiFan:
Want to add this to make sure that search engines are going to index the whole site.


dlech:
sounds good. We like robots, lol.

PR #76, opened 2015-11-29
Add structured data to improve looks in search results

WasabiFan:
I went through the structured data types that Google [say they accept](https://developers.google.com/structured-data/) and added all the applicable ones. Tested with their structured data analyzer tool. I think I replaced all the example content.

### DIFF
--- a/_includes/breadcrumbs.html
+++ b/_includes/breadcrumbs.html
@@ -1,12 +1,19 @@
 ﻿{% assign parent_dirs=page.url | replace_first: '/', '' | split: '/' %}
 {% assign link='/' %}
-<ol class="breadcrumb">
+<ol class="breadcrumb" vocab="http://schema.org" typeof="BreadcrumbList">
+    {% assign position = 1 %}
     {% for dir in parent_dirs %}
         {% if forloop.last %}
-            <li class="active">{{ dir }}</li>
-        {% else %}
-            {% assign link=link | append: dir | append: '/' %}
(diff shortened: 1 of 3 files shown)


dlech:
Thanks. I wonder if it is going to make a difference that we don't have breadcrumbs on every page.

_includes/breadcrumbs.html
```diff
@@ -1,12 +1,19 @@
 ﻿{% assign parent_dirs=page.url | replace_first: '/', '' | split: '/' %}
 {% assign link='/' %}
-<ol class="breadcrumb">
+<ol class="breadcrumb" vocab="http://schema.org" typeof="BreadcrumbList">
+    {% assign position = 1 %}
     {% for dir in parent_dirs %}
         {% if forloop.last %}
-            <li class="active">{{ dir }}</li>
```
dlech: I forgot to check before I merged this, so I don't know if this was a result of the changes here or the major changes you made previously, but in breadcrumbs, the active page appears as an active link in the breadcrumbs.

It looks like you aren't doing anything with the `class` variable in liquid after you are assigning it. I'm guessing that is what broke it.

WasabiFan: It looks like I partially reimplemented the classes but failed to carry it through -- I'll open a PR to fix this soon, or you can revert the merge of this PR and I can append it to this one.

dlech: new PR is just fine.

PR #77, opened 2015-12-01
Fix last link in breadcrumbs

WasabiFan:
See #76


dlech:
Very clever. I like it.

PR #80, opened 2015-12-04
Normalize and update table styles

WasabiFan:
See discussion in ev3dev/ev3dev#437.

This should include fixes for the issues pointed out in comments -- let me know if I missed anything.

I updated the sensor and motor pages to replace the `<td>`s with `<th>`s for the row headers that span multiple items. If there are any other pages that have similar row/column span settings in tables, we'll need to update those too to make them adopt the new style.

(Fixes ev3dev/ev3dev#437) -- this should close the other issue when you merge this.

### DIFF
--- a/_includes/motor.md
+++ b/_includes/motor.md
@@ -9,23 +9,23 @@
 
 ### General Info
 
-<table id="sensor-info">
-    <tr class="{% cycle 'info': 'd0', 'd1' %}">
+<table class="table table-striped table-bordered">
+    <tr>
         <th><code>device_name</code></th>
         <td><code>{{ motor.name }}</code></td>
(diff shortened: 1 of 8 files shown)


WasabiFan:
@dlech I just updated it so that it styles the first column separately. This required that I use an additional class for tables with this structure. I also added explicit `<thead>` and `<tbody>` tags to the tables which have this class. I might be able to remove the requirement for the explicit tags (I needed them for an earlier iteration of my trials but the new class might serve the necessary purpose), but don't have time to try it now.

Is this more in line with what you were expecting?

dlech:
Yeah, that's what I was going for. Thanks once again!

PR #83, opened 2015-12-09
Default to horizontal stacking for display of sysfs attributes

WasabiFan:
Closes ev3dev/ev3dev#449

Preview at <http://wasabifan.github.io/ev3dev.github.io/docs/drivers/tacho-motor-class/>

Also deleted useless file


dlech:
Nice. It's like a table without actually being at table.

Can we add a bit of space between each definition though?

I also think we should move the `(read/write)` part to the description, but that will have to be done in source code.

WasabiFan:
> Can we add a bit of space between each definition though?

Yes. I think. Uhhhh... actually I'm not quite sure. I'll get back to you on that. But probably.

> I also think we should move the  (read/write)  part to the description, but that will have to be done in source code.

I think that would work well too. And if you do that, I can remove the custom styling I added to widen the main column -- it's there so that `(read-only)` isn't converted to `(read-o...)` with the current text.

dlech:
OK. Some are still getting cut off, like `encoder_polarity`.

WasabiFan:
I just pushed some changes to add `10px` margin below each item -- do you think that that is enough, or should I make it something more like 15 or 20 pixels?

dlech:
There is no space after multiline items. I think the margin needs to go on the `dd` element, not `dt`. And I think 20px is better.

WasabiFan:
Updated.

dlech:
Awsome. much more readable now.

PR #87, opened 2015-12-30
Add new platform selection mechanism

WasabiFan:
Realization of ev3dev/ev3dev#474.

This is still WIP at the moment, so don't merge it (or look at it expecting it to be useful). But I have implemented the core of what was discussed in the issue linked above, so I figured I'd open this PR so that it can be discussed.

This just includes the basic JS needed to do the page content transformations and a test page that demonstrates the features. The code was originally a copy of the tab code, but as you can probably tell it has diverged to the point where there is very little resemblance (although the first 4 lines are _mostly_ unchanged :wink:). I also removed some functionality that I haven't yet reimplemented -- those items (and some others) are listed below. It also includes some notes-to-self and misc TODOs which will be cleaned up and removed in the future.

You can see the live test page [here](http://wasabifan.github.io/ev3dev.github.io/docs/tutorials/platform-test-page/).

Still unfinished:
- Nested attributes (for versions)
- Saving/loading previous selections
- Detecting platform options
- Formatting the options on the page with a flyout and whatever else
- Updating real pages

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -27,6 +27,7 @@
     <script src="{{ site.github.url }}/javascripts/search.js"></script>
     <!-- Converts static descriptors of tab target content into live text and links -->
     <script src="{{ site.github.url }}/javascripts/tabs.js"></script>
+    <script src="{{ site.github.url }}/javascripts/platform-select.js"></script>
     <!-- Adds classes and other styles to pages where hard-coded css can't be used -->
     <script src="{{ site.github.url }}/javascripts/style-helpers.js"></script>
 
--- a/docs/tutorials/platform-test-page.md
+++ b/docs/tutorials/platform-test-page.md
@@ -1,0 +1,37 @@
+---
+title: Platform test page
+subject: Website
+---
+
+This is a test page.
+
+*   {: data-platform-attribute-value="Android"}
+    1.  This is the first step for Android.
+
+    2.  This is the second step for Android.
+
+*   {: data-platform-attribute-value="Mac OSX"}
+
+    1.  Another step! This one's for Mac.
+
+    2.  ...And another for Mac.
+
+*   {: data-platform-attribute-value="Ubuntu"}
+
+    1.  Even more steps! Now featuring 100% more Ubuntu.
+
+    2.  Steps continued.
+
+*   {: data-platform-attribute-value="Windows"} I give up. Just imagine there are steps here.
+{: data-platform-select-list-attribute="client-platform" }
+
+*   {: data-platform-attribute-value="EV3 Brick"}
+    On the EV3, do a thing in this specific way.
+
+*   {: data-platform-attribute-value="BrickPi"}
+    The BrickPi is a little different. You actually need to do this other thing.
+
+*   {: data-platform-attribute-value="PiStorms"}
+    The PiStorms doesn't support this portion of the functionality.
+    You need to take this into consideration.
+{: data-platform-select-list-attribute="ev3dev-hardware-platform" }
--- a/javascripts/platform-select.js
+++ b/javascripts/platform-select.js
@@ -1,0 +1,118 @@
+function supportsHtml5Storage() {
+    try {
+        return 'localStorage' in window && window['localStorage'] !== null;
+    } catch (e) {
+        return false;
+    }
+}
+
+function cleanTextForId(text) {
+    return text.replace(/\W+/g, '-');
+}
+
+function getFilterByData(dataHash) {
+    if(arguments.length == 2) {
+        var newHash = {};
+        newHash[arguments[0]] = arguments[1];
+        dataHash = newHash;
+    }
+    
+    return function () {
+        var currentTarget = $(this);
+        return Object.keys(dataHash).every(function(property) {
+            return currentTarget.data(property) == dataHash[property];
+        });
+    }
+}
+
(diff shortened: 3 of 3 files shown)


dlech:
Looks like a good start.

One key feature I would like to have is to be like msdn where you have the tabs (or pills or whatever) to switch for *each* place on the page where there are platform specific bits rather than just one place to change it at the top of a page (unless of course the whole page is different).

For example, on [this page](https://msdn.microsoft.com/en-us/library/windows/apps/system.string.length%28v=vs.105%29.aspx), there are 3 different places where you can switch between C# and VB, but if you change one, not only does it change all 3 on that site, it also sets the default for the entire site until you change it again.

WasabiFan:
My plan is to do it as you mentioned, except have a single control on the page to control that (because it is changing it for the whole page anyway). More like a configurable setting.

So yes, my plan is that "not only does it change all 3 on that page, it also sets the default for the entire site until you change it again". Do you think it makes more sense to do it with the tabs instead of the page-level option? It basically comes down to whether or not you think of it as a global option to choose which type of documentation you are viewing or a localized option that changes a section of the docs that also saves it for convenience. The main idea that I see is that, because you are likely going to continue using the same platform for all your documentation browsing, we don't need to take up space in the content of the page for the options.

That MSDN page that you linked to does the "whole-page" thing for the .NET version (in the upper corner below the title) as well as the language selection. In terms of scale, the PC OS is probably something that changes the whole page while the ev3dev hardware only changes one or two details, so maybe I should make it so that you can set the mode (tabs or page-wide) in the markdown and have it render accordingly.

Whatever you want me to do here, it will only be a few small changes to make it serve either function.

WasabiFan:
By the way, this PR is not dead -- it's just deprioritized :wink: I have local changes that partially implement platform versioning, but I haven't had time to finish it up and push.

_One day..._

WasabiFan:
@dlech I have made some changes to this PR that get the platform selection fully-functional, barring the UI/UX. So, how would you like to have it laid out on the page?

I see two major options for this. One is to treat the platform as a "global preference":  the header has some sort of control that lets the user select their platform (the options would probably be in an options pane/flyout), and they see content that is appropriate for that platform. There wouldn't be a visual distinction between platform-specific content and generic content; to them, it would just be like a different page for each platform. This would likely include an icon to tell the user that their content is being modified for those preferences (which are accessible because the navbar is fixed in the viewport).

The second option is to have the platform control show up so that it is tied to each section of platform-specific content. In this model, the special content would be made visually distinct (maybe with a different background color?) and it would be presented to the user as a special section that could be changed to match their target. This would likely have the control either above or to the side of the section, and when the user clicks it it would expand into a full options pane.

We could also have a hybrid of these two modes; while the control is in the header, the platform-adapted content is still separated from the rest so that it is clear what content is different.

I think that this decision mainly comes down to how we intend to use the platform selection and how we want a reader to perceive it. For something like the getting started guide, for example, chances are that the platform-specific content is broken up into small chunks. In that case, having multiple selector controls in view at once may get confusing. On the other hand, if there is a large page with large blocks of platform text, it might be easier to understand if there is always a control clearly associated with adapted content.

---

Separately, we currently author the platform content in a `<ul>` and then have the tab logic pull the content out into a new `<div>`. Is there an advantage to this over just adding the divs from the start? I think that authoring the content as a set of divs has two main benefits: it reduces the amount of stuff that the script has to do, and it removes the need for careful indentation management when authoring the Markdown.

dlech:
> One is to treat the platform as a "global preference": the header has some sort of control that lets the user select their platform...

How would this work on mobile?

> The second option is to have the platform control show up so that it is tied to each section of platform-specific content...

This is more like the way I envisioned this working.

> I think that this decision mainly comes down to how we intend to use the platform selection and how we want a reader to perceive it...

This is pretty much how I plan on using it. Most of the content on all pages applies to all platforms. Then we have small sections that are platform-specific. The user should only see the platform they are interested in and the rest should be hidden so they don't have to scroll past irrelevant content.

> Is there an advantage to this over just adding the divs from the start?

If you can figure out a foolproof way to match divs to tabs, then great. Markdown requires careful indentation management no matter what you do, so I don't really see much of a benefit here. For me, the less HTML mixed in with the markdown the better.

WasabiFan:
> How would this work on mobile?

Theoretically, adding a special control to the header should let it flow just like other controls, although we might want to make it permanent on the nav bar instead of in the hamburger menu if we decide to go down that route. The other option is to have it sit at the top of the page so that you scroll to the top to access it.

> This is pretty much how I plan on using it. Most of the content on all pages applies to all platforms. Then we have small sections that are platform-specific.

I think that that format makes the case for the "whole-page" platform selection. If the content that changes is only small sections (but there may be a lot of them), having a control for each place that changes seems like it would probably be weird. To a typical user, seeing a bunch of the same UI on different parts of the page that are tied together would not be natural, and it would be especially awkward if each of those regions were colored separately from the rest of the page.

> The user should only see the platform they are interested in and the rest should be hidden so they don't have to scroll past irrelevant content.

Yep, I think we're on the same page there.

> If you can figure out a foolproof way to match divs to tabs, then great. Markdown requires careful indentation management no matter what you do, so I don't really see much of a benefit here. For me, the less HTML mixed in with the markdown the better.

Since it looks like you aren't sold on this idea, I'm going to leave it until I get to the point where I need to transfer pages over to the new selection system; then we can sort out how we want to do this.

PR #90, opened 2016-01-02
Better internal links

WasabiFan:
These changes replace the use of `site.github.url` with a new variable which is set in an include template. The template generates relative links, so we don't ever need to worry about where our stuff is: links will always resolve to the correct relative path.

This _would_ be a pretty elegant solution, but there's a caveat: while templates that are included into a page have access to variables declared on that page, content that is included in a layout does not. So this include needs to be on every content page where internal links are used.

This isn't great, but I think that it beats the alternative of either a) needing to manually edit URLs on dev machines because we always are redirected to the real site or b) being dependent on a specific URL structure that only works in some places. It _might_ be possible to do something like this with JS, but that brings with it many other issues and could be prone to errors.

Because I have changed so much stuff, **I would highly recommend not merging this directly**. Although I did my best to test key features such as search and the atom feed, I am around %90 sure that I broke and/or forgot something. So, I wanted to open this PR to make it easy to see just these changes; assuming @dlech is OK with the core of these modifications, I will squash and merge this branch into my CI branch so that it can run checks on every link on the site. Once I fix whatever errors show themselves, I expect that my CI changes will be happy. I will close this PR once we have come to a conclusion regarding usage of this new system.

Note to self: "Edit on GitHub" links should be updated to work in dev environment.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -12,6 +12,11 @@
     <link rel="stylesheet" href="{{ internal-link-base }}/stylesheets/bootstrap-extensions.css">
     <!-- Nav infrastructure and misc utils (search, user cards, embeds, CC license) -->
     <link rel="stylesheet" href="{{ internal-link-base }}/stylesheets/infrastructure-utils.css" />
+    
+    <!-- Page-specific variable that indicates the relative path to the site root -->
+    <script>
+        var pageLinkBaseUrl = "{{internal-link-base}}";
+    </script>
 
     <!-- 3rd-party libraries -->
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.0/jquery.min.js"></script>
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -4,31 +4,31 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
     <title>{{ page.title }}</title>
-    <link rel="alternate" type="application/atom+xml" href="{{ site.github.url }}/news/atom.xml" title="Atom feed">
+    <link rel="alternate" type="application/atom+xml" href="{{ internal-link-base }}/news/atom.xml" title="Atom feed">
     <link rel="icon" href="favicon.ico" />
 
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
-    <link rel="stylesheet" href="{{ site.github.url }}/stylesheets/bootstrap.css">
-    <link rel="stylesheet" href="{{ site.github.url }}/stylesheets/bootstrap-extensions.css">
+    <link rel="stylesheet" href="{{ internal-link-base }}/stylesheets/bootstrap.css">
+    <link rel="stylesheet" href="{{ internal-link-base }}/stylesheets/bootstrap-extensions.css">
     <!-- Nav infrastructure and misc utils (search, user cards, embeds, CC license) -->
-    <link rel="stylesheet" href="{{ site.github.url }}/stylesheets/infrastructure-utils.css" />
+    <link rel="stylesheet" href="{{ internal-link-base }}/stylesheets/infrastructure-utils.css" />
 
     <!-- 3rd-party libraries -->
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.0/jquery.min.js"></script>
     <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/js/bootstrap.min.js"></script>
-    <script src="{{ site.github.url }}/javascripts/jquery.getUrlParam.js"></script>
-    <script src="{{ site.github.url }}/javascripts/jquery.loadTemplate-1.4.4.min.js"></script>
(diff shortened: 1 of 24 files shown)


WasabiFan:
@dlech can you "OK" these changes or express your dissent? If you're OK with handling paths like this, I will merge this into my CI branch to make most of the warnings shut up.

dlech:
Sorry, I've been on holiday for the past couple days. What exactly is the problem we are trying to solve here? On my dev machine, links work without sending me to the "real" site. I don't understand why we need to do this.

WasabiFan:
The way we currently do it is hacky and will break if we want to use GitHub metadata on the site, and can't work anywhere other than if it is built and hosted on GH Pages. The way that I implemented here guarantees that we will never need to deal with link issues due to environment changes or feature additions, so I'd say it is better to embrace the changes now instead of discovering issues later.

dlech:
I seem to be rather dense today, but I am finally getting my head around this. Basically, the issue we are trying to solve is <https://github.com/jekyll/jekyll/issues/332>.

Since we are using our own domain name (cname), we actually don't have a problem between local development and the live website. If we were to remove all of the instances of `{{ site.github.url }}` completely, things are still working nicely. The only problem is that if you want to publish this to your personal `gh-pages` branch to make a public preview of your changes, then that is when we have problems because of the added `basename` directory (i.e. your site's root is at `https://user.github.io/ev3dev.github.io/` instead of just `https://user.github.io/`).

Since the only problem is with previewing, why not do something like [this](https://github.com/jekyll/jekyll/issues/332#issuecomment-104685178) instead? Except that instead of using `{{ site.baseurl }}` everywhere in the source, the rake file should search all generated files for links that start with `/` and prepend `/ev3dev.github.io` to them before committing. This would actually make generating previews easier than trying to wrangle feature branches with the `gh-pages` branch and it would also eliminate the annoying emails about `cname`.

WasabiFan:
> Basically, the issue we are trying to solve is jekyll/jekyll#332.

Yes, that is what we are trying to do as far as I am concerned.

> The only problem is that...

The other issue that I am having is that root-relative paths (e.g. `/foo/bar

WasabiFan:
> Basically, the issue we are trying to solve is jekyll/jekyll#332.

Yes, that is what we are trying to do as far as I am concerned. But there is a (small) added wrinkle; see below.

> The only problem is that...

The other issue that I am having is that root-relative paths (e.g. `/foo/bar`) only work via protocols and servers that use the site root as the URL root (this is the root cause of the forked page previews issue). Test scripts on the CI server can't handle root-relative paths, because it is exploring the file structure relative to the system root, not the site root. So the path `/foo` on a local machine just resolves to the root of the system drive, not the site folder.

> Since the only problem is with previewing, why not do something like this instead? Except that instead of using  {{ site.baseurl }}  everywhere in the source, the rake file should...

So, the proposal is to stop using GitHub Pages to build and just use it as a static file host? If we did that, people would no longer be able to push to their `gh-pages` branch to see the site. And we would need to set up our own build server. But technology-wise, it should work.

dlech:
> So, the proposal is to stop using GitHub Pages to build and just use it as a static file host? If we did that, people would no longer be able to push to their gh-pages branch to see the site.

Actually, I'm saying the opposite. People should push the generated files (just the `_site` directory) to their `gh-pages` branch for preview. There would be a script that you run that does this automatically. We would continue to do local development and deployment as we currently do.

> The other issue that I am having...

This same script could actually solve this problem as well. For previewing `gh-pages` branches, the script would prepend `/ev3dev.github.io` to all root relative paths. For the build server, on the other hand, the same script could prepend `/local/path` (whatever it really is) to the root relative paths.

WasabiFan:
>  People should push the generated files (just the  _site  directory) to their  gh-pages  branch for preview. There would be a script that you run that does this automatically.

OK, I think I understand now. So we do it as we currently do both in code and procedure, but write a script that we use for deployment to test branches and on the CI server that replaces links that start with `/` with a better local alternative. How would this work if we decide we want to use the `GitHub-pages` plugin to get GitHub metadata? Then it would start filling `site.github.url` with the remote path on dev machines as well. Would we need to write a plugin that overwrote that property?

Going back to an earlier comment, how does doing it this way "... make generating previews easier than trying to wrangle feature branches with the  gh-pages  branch"? If I am understanding the proposal correctly, we would still only have one personal preview available at a time and publishing from one branch would overwrite the changes from the other.

dlech:
> Would we need to write a plugin that overwrote  [the `site.github.url`] property?

I  would say just never use the `site.github.url` property. We wouldn't need it anymore.


> If I am understanding the proposal correctly, we would still only have one personal preview available at a time and publishing from one branch would overwrite the changes from the other.

yes.

WasabiFan:
> I would say just never use the  site.github.url  property. We wouldn't need it anymore.

Ahhh I see. Cool!

I can try to implement this when I get the chance.

dlech:
I'm tinkering with it right now, so hopefully you haven't started.

WasabiFan:
> I'm tinkering with it right now, so hopefully you haven't started.

I have not -- I have enough other stuff to do at the moment :wink:

PR #91, opened 2016-01-04
Get rid of {{ site.github.url }}

dlech:
* Also get rid of absolute links to http://www.ev3dev.org
* Introduce publish.rb script to publish to personal gh-pages branch
  (this was the whole reason for using site.github.url in the first place)

Fixes #90

### DIFF
--- a/news/index.html
+++ b/news/index.html
@@ -12,7 +12,7 @@ subtitle: Announcements and updates
 <div class="panel panel-default">
     <div class="panel-heading">
         <h1 class="panel-title">
-            <a href="{{ site.github.url }}{{ first_post.url }}"> {{ first_post.title }}</a>
+            <a href="{{ first_post.url }}"> {{ first_post.title }}</a>
             <span class="pull-right label font-light">{{ first_post.date | date_to_string }}</span>
         </h1>
     </div>
@@ -20,7 +20,7 @@ subtitle: Announcements and updates
         {{ first_post.content | truncatewords: 200 }}
         <br/>
         <br />
-        <a href="{{ site.github.url }}{{ first_post.url }}">Continue reading...</a>
+        <a href="{{ first_post.url }}">Continue reading...</a>
     </div>
 </div>
 
@@ -32,7 +32,7 @@ subtitle: Announcements and updates
             <div class="panel panel-default">
                 <div class="panel-heading">
                     <h1 class="panel-title">
-                        <a href="{{ site.github.url }}{{ post.url }}"> {{ post.title }}</a>
+                        <a href="{{ post.url }}"> {{ post.title }}</a>
                         <span class="pull-right label font-light">{{ post.date | date_to_string }}</span>
                     </h1>
                 </div>
@@ -40,7 +40,7 @@ subtitle: Announcements and updates
                     {{ post.content | truncatewords: 50 }}
                     <br />
                     <br />
-                    <a href="{{ site.github.url }}{{ post.url }}">Continue reading...</a>
+                    <a href="{{ post.url }}">Continue reading...</a>
                 </div>
             </div>
         {% endif %}
@@ -48,5 +48,5 @@ subtitle: Announcements and updates
 {% endfor %}
 
 <div class="row text-center">
-    <a class="btn btn-primary" href="{{ site.github.url }}/news/archive">See news archive for older posts</a>
+    <a class="btn btn-primary" href="/news/archive">See news archive for older posts</a>
 </div>
--- a/publish.rb
+++ b/publish.rb
@@ -1,0 +1,59 @@
+#!/usr/bin/env ruby
+#
+# publish.rb
+#
+# This script publishes the ev3dev website to the gh-pages branch of your
+# personal fork. This allows you to share a preview of your changes with others.
+#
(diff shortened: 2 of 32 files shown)


dlech:
@WasabiFan, take this for a spin. You can see the live preview (generated by `./publish.rb dlech`) at <http://dlech.github.io/ev3dev.github.io/>.

WasabiFan:
Search is broken; make sure to update that as well. The search code uses URLs in two places: in `search.js` it loads the search index, and in the search index JSON file it sets the URLs to go to. You can look at the branch from my previous PR to see how you can send relative links in the JSON and modify them in the JS if you want to do it that way.

news/index.html
```diff
@@ -12,7 +12,7 @@
 <div class="panel panel-default">
     <div class="panel-heading">
         <h1 class="panel-title">
-            <a href="{{ site.github.url }}{{ first_post.url }}"> {{ first_post.title }}</a>
+            <a href="screenshot.html source="{{ site.github.url }}/{{ first_post.url }}"> {{ first_post.title }}</a>
```
WasabiFan: What is this?

dlech:
> Search is broken

Yup, there if a TODO in publish.rb for that :wink:

news/index.html
dlech: No idea.

WasabiFan: Well that's a good sign :wink:

WasabiFan:
> there if a TODO in publish.rb for that :wink:

I haven't gotten that far in the diff yet :wink: I'm multitasking

news/index.html
dlech: I cleaned up this mess and force pushed, so should be better now.

WasabiFan:
It doesn't look like the atom feed was changed; you may or may not want to change that, depending on the behavior you're looking for.

repo-metadata.html
```diff
@@ -2,18 +2,7 @@
...
-        <section>
-            {% include title.html %}
-            <strong>site.github</strong>
-            {{ site.github }}
-        </section>
-        {% include footer.html %}
-    </div>
-</body>
-</html>
+{% for x in site.github %}
```
WasabiFan: You could do `{{ site.github | jsonify }}` instead here to get the whole hierarchy

publish.rb
```diff
@@ -0,0 +1,59 @@
...
+    STDERR.puts "Usage: ./publish.rb <gh-user-name>"
+    exit(1)
+end
+
+Dir.mktmpdir do |tmp|
+    FileUtils.cp_r "_site/.", tmp
+    Dir.chdir tmp
+    system "git init"
+    git_url = "git@github.com:#{ARGV[0]}/ev3dev.github.io.git"
+    system "git remote add origin #{git_url}"
```
WasabiFan: What if I already have a remote configured that points to my fork, or already have something named `origin` configured? And this should use `https` instead so that it works on Windows.

dlech:
I figured that the atom feed needed absolute urls, so I left them. And I figured that it didn't matter if it work anywhere else. If it ever gives us trouble, github recently allowed [jekyll-feed](https://help.github.com/articles/atom-rss-feeds-for-github-pages/). But, if it ain't broke, don't fix it.

publish.rb
```diff
@@ -0,0 +1,59 @@
...
+    # $.getJSON("/search-index.json", function (e) {
+
+    # TODO: need to fixup search-index.json
+    # line like this need to be changed...
+    # "href": "/news/2015/12/30/ev3dev-jessie-2015-12-30-release/",
+
+    system "git add ."
+    message = "Site updated at #{Time.now.utc}"
+    system "git commit -m #{message.inspect}"
+    system "git push origin master:refs/heads/gh-pages --force"
```
WasabiFan: Looks like this assumes that you're on `master`... it should check the current branch instead I suppose

publish.rb
dlech: This creates a new local git repository in a temporary directory, so it can't conflict with anything.

I always use ssh for git, even on Windows. I don't like using https because you have to enter your password every time. I would be ok adding a command line switch so that you can choose one or the other.

repo-metadata.html
dlech: Cool. This file is not actually used at the moment though.

publish.rb
dlech: again, this is starting with a new empty git repository in a temporary directory. Assuming `master` is 100% safe unless there is a way to locally configure git to use another name for the default branch.

publish.rb
WasabiFan: Command-line switch would be appreciated.

WasabiFan:
Looks good. Once you merge this, I can update my CI branch from master and modify the script to work for my needs there. If you have already configured command-line parsing at that point, the changes should be fairly trivial.

Although, now that I think about it, maybe this script shouldn't publish by default -- it would just create the files. Then you could add `--publish wasabifan` to publish to a repo. Otherwise I can add a `--no-publish` flag to use on the CI server.

WasabiFan:
Oh, and we will need to update the project publishing tutorial for these changes; you just won't be able to preview your changes unless you want to do it on the command line.

PR #100, opened 2016-01-16
href="#" to fix drop down menus

moriarty:
proposed fix for drop down menu not appearing in narrow desktop browsers. See ev3dev/ev3dev#480.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -17,7 +17,7 @@
                         <a href="/" title="There's no place like Home...">Home</a>
                     </li>
                     <li class="dropdown">
-                        <a class="dropdown-toggle" href="#main-nav" data-toggle="dropdown" title="Documentation on using ev3dev">
+                        <a class="dropdown-toggle" href="#" data-toggle="dropdown" title="Documentation on using ev3dev">
                             Docs
                             <span class="caret"></span>
                         </a>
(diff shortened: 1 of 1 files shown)


dlech:
Thanks!

WasabiFan:
Uhhh... This makes the site HTML invalid. Linking to `#` is linking to something that doesn't exist, and that is considered invalid by the HTML gods (a.k.a. The standards people). You _could_ tell our proofer to ignore invalid links that point to `#`, but I think it would be best to find a different way.

dlech:
Invalid is better than broken, yes? Also, html-proofer has an option for specifically ignoring this because it is a very common pattern (which I turned on).

moriarty:
@WasabiFan , what about removing the href altogether? I saw that it was ```href="#"``` in all of the examples using the same theme, so that's how I thought to fix it. 

http://stackoverflow.com/questions/10510191/valid-to-use-a-anchor-tag-without-href-attribute

I tried this in the google chrome inspector and it also appears to solve the problem.

WasabiFan:
> what about removing the href altogether? I saw that it was  href="#"  in all of the examples using the same theme, so that's how I thought to fix 

All HTML `<a/>` tags are supposed to have an `href` :wink: But as @dlech said, at some point we just have to deviate from the rules and ignore it. It seems like our best option is just to use nonexistent `#` links, but it's not optimal.

moriarty:
Yes, I thought it was suppose to have an ```href``` also, but apparently it's still valid without one. I should have been more specific with my reference, not the question, but the best answer to that stackoverflow

https://www.w3.org/html/wg/drafts/html/master/semantics.html#the-a-element

WasabiFan:
> Yes, I thought it was suppose to have an  href  also, but apparently it's still valid without one. 

Looks like it is; I was just going off of the HTML validators that I have used previously, and those complain when you're missing a link. If it's not required, we can do it however seems easiest -- which in this case seems to be to just leave it as it is.

PR #105, opened 2016-02-11
Switch info alert to be an info panel on brickstrap page

WasabiFan:
Putting an essay in an alert makes it very big :wink:

These changes were made in the online editor, so we should check this after merging.


WasabiFan:
I don't think we have icons in other panels, so I normalized it -- if we include it here we should do the same wherever else we have panels.

dlech:
I think I added icons to most panels.

PR #109, opened 2016-02-27
PIX3L PLOTT3R project added

droidsrobotics:
We added a project page for the PIX3L PLOTT3R.  Hope it looks right.  The content for the page and the YouTube video are done. We are still in the process of finishing the LDD for the design. We will add that soon, but there are pictures for the build on the project’s homepage.  Code has been shared using github. We will keep that updated.

### DIFF
--- a/projects/_posts/2016-02-26-PIX3L-PLOTT3R.md
+++ b/projects/_posts/2016-02-26-PIX3L-PLOTT3R.md
@@ -1,0 +1,21 @@
+---
+author: [ "@droidsrobotics", "Sanjay and Arvind Seshan" ]
+programming_language: "Python" 
+
+youtube_video_id: rrNaLfE9PWA # The video ID of the YouTube video to be displayed with this post
+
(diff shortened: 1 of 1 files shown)


WasabiFan:
This is really cool! I think that's the best resolution I've seen on a LEGO-based printer of any sort.

droidsrobotics:
Don't know why the checks are failing -- this is our first time with this type of pull request. It seems to be complaining about http://honk.sigxcpu.org/projects/git-buildpackage/manual-html/gbp.html. However, this is not something we changed.

dlech:
It looks like there is a link broken somewhere else on the site that is causing the failure, so nothing to worry about here.

I've been following your progress of Facebook. Thanks for sharing here too.

dlech:
http://www.ev3dev.org/projects/2016/02/26/PIX3L-PLOTT3R/

rhempel:
Great job guys!

PR #116, opened 2016-02-27
Use Sass for Bootstrap, Bootswatch and custom styles

### LINKED ISSUE #111: Switch website to using Sass instead of plain CSS
[Jekyll natively supports Sass](http://jekyllrb.com/docs/assets/). Instead of using custom-compiled Bootstrap and CSS additions, we should see if we can get Jekyll to build Bootstrap for us. Then our custom theme modifications won't need to be manually compiled, and we can stop using CSS hacks to modify Bootstrap code.

------------------

I have a successful local test of building Bootstrap + Bootswatch in a Jekyll site. At some point soon I'll try to get our custom adaptations extracted from our site, re-written as Sass, and then put in the right place so that Bootstrap builds with our custom stuff.

From my tests, it looks like this adds around a second of build time; this isn't optimal, but hopefully with Jekyll 3.0's new profiling tool we can speed up our site build anyway.

WasabiFan:
This replaces our precompiled Bootstrap and Bootswatch files with Sass that is compiled on-the-fly. This means that our customizations are more modular and that updating upstream versions is as simple as copying some files and making sure that the thing compiles successfully.

I have confirmed that this builds correctly in safe mode, but because of the relative link issue on personal pages I cannot test it for sure on GitHub Pages. I did, however, publish this to my personal fork, have GitHub build it, and then load the page and manually fix the header links. As far as I can tell, it does work on GitHub's infrastructure. There are no visual or functional differences when comparing the real site and this modified version.

I also added a markdown document with some notes on how this works, so that others can understand what is going on without me having to be available. I hope to add more to this document as time goes on.

Fixes #111.


dlech:
> but because of the relative link issue on personal pages I cannot test it for sure on GitHub Pages

We fixed this. Does `./publish.rb` not work for you?

_includes/header.html
```diff
@@ -106,7 +106,4 @@
             </div>
         </div>
     </nav>
-</div>
-
-<!-- make sure this get included on every page -->
-<!--[if !IE]><script>fixScale(document);</script><![endif]-->
+</div>
```
dlech: newline at end of file. Since this is an include, it could have bad effects if it is missing.

WasabiFan:
I know that we have the script, but that doesn't test whether GitHub has any security settings that prevent it from building Sass. The script just publishes locally-built files for viewing online, which isn't very useful in this case because there aren't any visual changes to see.

_includes/header.html
WasabiFan: Will-do.

dlech:
Oh, right

WasabiFan:
Changes made.

PR #124, opened 2016-03-10
Updated all links and youtube video

droidsrobotics:
We created a new YouTube video that includes all the newest features (including a user interface). So we updated the link.

We also updated both the source code and building links since we decided to add more projects (not just this ev3dev one) to our github and had to reorganize a little bit.


WasabiFan:
Err, from my quick look I think that the CI build failure is our fault, not yours. @dlech and I will need to figure that out. Sorry for any potential confusion!

PR #132, opened 2016-04-10
Ported NFS tutorial and some more stuff

antonvh:
The ev3 complained about some statd service not running. Therefore the only way to mount a share was `sudo mount -o nolock ~/nfs/mac` Is there a better way?


WasabiFan:
Although most of the CI errors are our issues (not yours), there are a few legitimate broken links that it is pointing out:

- ./docs/tutorials/setting-up-an-nfs-fileShare/index.html
  *  `[http://forum.stmlabs.com/showthread.php?tid=6285` is an invalid URL (line 305)
  *  internal image Images/Website/nfs_on_ev3dev.png does not exist (line 257)

_includes/title.html
```diff
@@ -19,10 +19,14 @@
...
         {{ page.title | xml_escape }}
         <small>{{ page.subtitle }}</small>
+        
         {% if edit_path %}
             <a class="btn btn-primary pull-right" href="{{ edit_path }}">Edit on Github</a>
         {% elsif show_news_link %}
             <a class="btn btn-primary pull-right" href="/news/atom.xml">Subscribe via Atom</a>
         {% endif %}
     </h1>
+    {% if page.author %}
```
WasabiFan: This title is included on every page, so this will have unintended consequences. Info like this should go in the page- or category-specific templates.

Separately, I'm not sure that we want a visible author tag on tutorials. While attribution on the project pages is pointing out whose work it is (because they still own the rights to it), in a tutorial as I see it the purpose is more to tell future contributors who to talk to if they have questions. I think that previous discussions that we have had concluded that content everywhere but the project pages was the property of the ev3dev organization and not the original author(s).

This is just my opinion and it's not a strong one, so I think @dlech will have to make a decision that we stick with from now on.

docs/tutorials/setting-up-an-nfs-fileShare.md
```diff
@@ -0,0 +1,180 @@
...
+subject: Networking
+author: Anton Vanhoucke, Ralph Hempel
+---
+
+* Table of Contents
+{:toc}
+
+## Why Do It?
+
+The nfs protocol is the standard Linux way to share files between computers, and since you're running Linux on the EV3, and there's nfs support built-in, you can share files between your computer and the EV3.
```
WasabiFan: NFS is an acronym, so should be capitalized

docs/tutorials/setting-up-an-nfs-fileShare.md
```diff
@@ -0,0 +1,180 @@
...
+
+Using a file share method like nfs makes it easier for folks like me to load and unload modules that are under development. That saves time and is less error prone than swapping SD Cards and rebooting all the time.
+
+For general users, it makes it easy to upgrade their SD Cards with known good modules, new kernel `uImage` files, and other updates to config files. It also makes it easier to backup your SD card. Finally, it allows you to edit files on your host machine, then run them on the EV3 without having to actually copy files anywhere!
```
WasabiFan: I think that the "edit files then run them" bit is the most important for the majority of users, so it might make sense to lead with that.

docs/tutorials/setting-up-an-nfs-fileShare.md
```diff
@@ -0,0 +1,180 @@
...
+It's relatively easy, you just need to keep the addresses and filenames straight. All you have to do is modify one file on the EV3 and one on your host computer. Let's start with getting the computer set up.
+
+The instructions here are for all three platforms Linux, Windows, and OSX - just pick the one you need 
+
+## How To Do It - Linux
+
+On your Linux box, you'll need to edit a file called `/etc/exports`. If you don't have this file, then you need to install `nfs-common` or a similar package. Update the file to look like this:
+
+```
+# /etc/exports: the access control list for filesystems which may be exported
```
WasabiFan: Assuming that these comments show up in this file by default, I think it would be clearer if the comments were omitted and it just said "... Open the file and add the following line:".

docs/tutorials/setting-up-an-nfs-fileShare.md
```diff
@@ -0,0 +1,180 @@
...
+- Disables subtree checks, more reliable but mild security risk
+- Do not let root on the EV3 to be like root on the host, good idea!
+
+So update the file on your host machine, then run `sudo exportfs -rv` which will update the directories that nfs exports.
+
+Now update the fstab on the EV3 (see below). 
+
+## How To Do It - Windows
+
+We're in luck! You can download [haneWin NFS](http://www.hanewin.net/nfs-e.htm) which is an nfs server that works on Windows 200x/XP/Vista/7 - hopefully on Windows 8. I have it running on Windows 7 and it's great. The haneWin server has built-in help for setting up the Windows side of the share, and if you get stuck there are [other users on the net that have it working][stmlabshanewin].
```
WasabiFan: It would be helpful to add a note at the top of the section (move the existing one) that warns that this is paid software. Personally, I would recommend against people using this software on Windows because it is a) paid and b) not particularly well-maintained. Having said that, it is a valuable option to document and I am sure that some people will want to use it, so It should still be in here. 

Note that you can add an "alert" on the site using the following Markdown, which should make it more visually apparent:
```
{% include icon.html type="warning" %}
Your text goes here (and <https://example.com>{: .alert-link} links need that extra alert-link notation).
{: .alert .alert-warning}
```

docs/tutorials/setting-up-an-nfs-fileShare.md
```diff
@@ -0,0 +1,180 @@
+---
+title: Setting Up an NFS Fileshare
+subject: Networking
+author: Anton Vanhoucke, Ralph Hempel
+---
+
+* Table of Contents
+{:toc}
+
+## Why Do It?
```
WasabiFan: I think that there should be a notice at the top saying that this is generally for advanced users. We don't want people who aren't prepared to do a some manual configuring getting stuck here.

You can use an info-style alert for this:
```
{% include icon.html type="info" %}
This is your info text
{: .alert .alert-info}
```

dlech:
> Is there a better way?

Does this help? http://unix.stackexchange.com/questions/77055/mount-linux-nfs-rpc-statd-is-not-running

dlech:
We have 2 nfs related services disabled by default: https://github.com/ev3dev/brickstrap/blob/master/projects/ev3dev-jessie/base/hooks/systemd

Have you enabled them both?

    systemctl enable nfs-common.service
    systemctl start nfs-common.service
    systemctl disable rpcbind.service
    systemctl start rpcbind.service

_includes/title.html
antonvh: I personally find that authorship is good for two reasons:
- people can find you on github for questions. Tutorials will probably expire some day. And I won't know until someone asks me.
- I think it is motivating. It is some kind of recognition. You see it a lot in the modding community. Every contributor's two minutes of fame. :) I was thankful to @rhempel that he mentioned me in the ev3dev-lang-python source.

I believe more (potential) contributors are like me in the above cases. Let's motivate them!

WasabiFan: Hmmm, that makes sense -- thanks for the explanation! I'd like to look into ways to prominently display the contributor names in a way that says "this person contributed" but still keeps it communal and open. I have some ideas that I'll flesh out when I get the chance, but the changes I want to make are mainly just presentation of this info. Maybe a "special thanks" section or similar. The point on contactability is also an important one, so I might try to use the GitHub user cards.

Anyway, for now I think this is fine. Just make sure that you modify it so that this line only shows up in tutorials, not projects or other pages.

dlech:
Related: https://github.com/ev3dev/ev3dev/issues/85

_includes/title.html
antonvh: Also: there is no separate tutorial layout. So it's hard to restrict this to the tutorial pages. The {% if %} statement at least makes sure there is no ghost <p> when the page has no author property. Github cards would be much better btw! Maybe they can be inserted as a block, like a toc?

WasabiFan: I see the issue. Currently, I expect that this addition will cause this to show up on project pages too (where we already use an `author`property), which would be duplicating information. For now, can you just rename this to `tutorial_author`, `docs_author` or similar so we avoid the collision?

As for the box, have you taken a look at the project pages? We already have that info box functional if we'd like it here. The problem is that it takes up a fair amount of horizontal space. So I'd say leave it as-is for now, and when I get the chance I can try to modify it to make it work here without taking up too much space.

dlech: Why not just just create a new layout? For example, it should be fairly trivial to copy `_layouts/docs.html` and replace the contents with an author `<div>`. You just have to change <https://github.com/ev3dev/ev3dev.github.io/blob/master/_config.yml#L36> from `pages` to `tutorial` or whatever you name the new layout.

dlech:
Thanks!

PR #138, opened 2016-04-23
fix news index page

### LINKED ISSUE #137: Latest failure.
Don't know what this means and don't feel like looking into it at the moment, so I'll leave it here for investigation later...

```
- ./news/index.html
  *  Tag code... invalid (line 201)
  *  error parsing attribute name (line 201)
htmlproofer 3.0.5 | Error:  HTML-Proofer found 2 failures!
```

dlech:
using truncate words was not good because it would truncate in the
middle of html tags and cause errors.

fixes #137 

The way this works is that it shows whatever if before the first heading. This isn't ideal for some of the existing posts, but going forward, we can always make sure to have a good introduction before the first heading.

### DIFF
--- a/news/index.html
+++ b/news/index.html
@@ -3,30 +3,9 @@ title: News
 subtitle: Announcements and updates
 ---
 
-{% for post in site.posts reversed %}
-    {% if post.categories contains "news" %}
-        {% assign first_post = post %}
-    {% endif %}
-{% endfor %}
-
-<div class="panel panel-default">
-    <div class="panel-heading">
-        <h1 class="panel-title">
-            <a href="{{ first_post.url }}"> {{ first_post.title }}</a>
-            <span class="pull-right label font-light">{{ first_post.date | date_to_string }}</span>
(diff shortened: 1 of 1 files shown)


WasabiFan:
Hmmm... I was pretty sure I fixed that. Maybe it was on a different index page. But whatever the case, instead of finding the heading, I would recommend using the strip HTML filter (can't remember the name) before truncating.

WasabiFan:
Oh, I remember -- I think I did that on the projects page.

dlech:
I tried stripping the html and it was really ugly. I like this way better.

dlech:
preview: http://dlech.github.io/ev3dev.github.io/news/

WasabiFan:
Oh, I see -- I guess there's a lot more HTML in the news pages than the project pages. Yeah, I like your solution better.

dlech:
What the heck! It looks different on the actual site compressed to my preview!

dlech:
all better

PR #140, opened 2016-04-24
Fix floating user cards and update styles for current usage

### LINKED ISSUE #139: github-cards need tweaking for horizontal layout
Example from http://www.ev3dev.org/docs/tutorials/setting-up-an-nfs-file-share/

selection_002

WasabiFan:
This does two main things:
- It fixes the width issue when the card is "floating" at the bottom of the tutorial pages (fixes #139)
- It adds some additional styling to make the cards more visible on the normal page background, as well as look better overall

Unfortunately, I can't publish a preview at the moment because the publish script explodes on Windows and the bash shell gets very unhappy when I try to install Ruby. I'm looking for a fix to the latter issue.

### DIFF
--- a/_layouts/tutorial.html
+++ b/_layouts/tutorial.html
@@ -26,6 +26,6 @@ layout: page
 <h2> Author </h2>
 {% endif %}
 {% for author in page.author %}
-<div style="float: left; margin-right: 20px;">{% include author-card.html author = author %}</div>
+<div style="float: left; margin-right: 20px;">{% include author-card.html author=author theme='light' %}</div>
 {% endfor %}
 {% endif %}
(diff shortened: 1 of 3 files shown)


_layouts/tutorial.html
```diff
@@ -26,6 +26,6 @@
 <h2> Author </h2>
 {% endif %}
 {% for author in page.author %}
-<div style="float: left; margin-right: 20px;">{% include author-card.html author = author %}</div>
+<div style="float: left; margin-right: 20px;">{% include author-card.html author=author theme='light' %}</div>
```
dlech: Why make this a different "theme" from other user cards?

WasabiFan: Because the background color in the info panel on other pages is lighter than the page background here. The drop shadow didn't give it enough contrast in my opinion to make it easily visible (and shadows don't fit with the other visuals on our site anymore), so I made the card background lighter on dark backgrounds and darker on light backgrounds. The themed highlight on the left side also helps with the contrast.

I called it "theme" just so that it is easy to add other context-specific styles if needed; for example, it might make sense to have a large version and a small version at some point.

PR #145, opened 2016-04-27
Add meta descriptions with proper content on important pages

WasabiFan:
This adds logic to the head template that generates a description `meta` tag with content from the page's `excerpt`. That field is automatically populated on posts, and you can manually override or add it on any page you want. The description content is often what search engines display on results pages, so we want to make sure we have good content there.

The majority of the content is copied from the first paragraph on the page, but in some cases I added an extra sentence or modified wording to make it make sense in search results. On some of these, I was having some trouble coming up with good descriptions -- if there are any suggestions for content, I'd be happy to make changes.

I focused on the primary landing pages and the major posts and documentation items, but didn't add excerpts for everything. It would also be great if the script that generates docs pages would add excerpts.

### DIFF
--- a/docs/ports/index.md
+++ b/docs/ports/index.md
@@ -1,6 +1,6 @@
 ---
 title: Input and Output Ports
-excerpt: "Documentation reference for the ports that one plugs sensors and motors into on the EV3 and their associated ev3dev drivers."
+excerpt: "Documentation reference for the input (sensor) and output (motor) port device drivers."
 ---
 
 This page is about the kind of ports that you plug [sensors] and [motors] into.
--- a/docs/ports/index.md
+++ b/docs/ports/index.md
@@ -1,5 +1,6 @@
 ---
 title: Input and Output Ports
+excerpt: "Documentation reference for the ports that one plugs sensors and motors into on the EV3 and their associated ev3dev drivers."
 ---
 
 This page is about the kind of ports that you plug [sensors] and [motors] into.
--- a/share.md
+++ b/share.md
@@ -1,5 +1,6 @@
 ---
 title: Share
+excerpt: "Do you like ev3dev? Show your support! We invite you to star the repo on GitHub and submit your cool projects to our projects page."
 ---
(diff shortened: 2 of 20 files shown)


docs/ports/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Input and Output Ports
+excerpt: "Documentation reference for the ports that one plugs sensors and motors into on the EV3 and their associated ev3dev drivers."
```
dlech: This wording is a bit awkward to me. How about "Documentation reference for the input (sensor) and output (motor) port device drivers."

share.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Share
+excerpt: "Do you like ev3dev? Show your support! We invite you to star the repo on GitHub and submit your cool projects to our projects page."
```
dlech: This isn't the best choice for an excerpt. It actually would be good as an alert on this page. The excerpt should say something to the effect that this page lets you find project that other have shared and tells you how to share your own projects.

WasabiFan: Ummm... But that's not what this page does. Are you saying that that's what it _should_ be?

docs/ports/index.md
WasabiFan: That sounds better. I'll update it.

share.md
dlech: I guess I should have looked at the page first :blush: 

Still, I think the except should come from the Projects rather than the other stuff on the page.

WasabiFan:
@dlech Changes made. It would be great if we had longer text for some of these (otherwise they will have to be automatically augmented by search engines), but this should be fine for now.

dlech:
Well, let's give it a try and see what happens.

WasabiFan:
> Well, let's give it a try and see what happens.

Are you saying that this should be merged? Or was there something else I should do?

dlech:
Yes, I thought I merged it already. Let me try that again...

PR #156, opened 2016-05-12
Update NXT Temperature Sensor entry

kortschak:
Waiting on ev3dev/lego-linux-drivers#32.


dlech:
Thanks for doing this. There is actually another bit that can be removed [here](https://github.com/ev3dev/ev3dev.github.io/blob/master/_includes/sensor-connection.html#L33).

Note to self: don't merge this until kernel cycle 12 release.

kortschak:
Done.

kortschak:
Sorry, there's limited information in the branches page to show whether branches are merged between forks. I was overzealous.

PR #157, opened 2016-05-13
Create supported-wi-fi-adapters

JorgePe:
A list of Wi-Fi USB adapters known to work with ev3dev
Do markup works in this files? I tried a list and a block of command-line text but doesn't show at preview.

### DIFF
--- a/docs/tutorials/supported-wi-fi-adapters
+++ b/docs/tutorials/supported-wi-fi-adapters
@@ -5,7 +5,7 @@ subject: Networking
 
 LEGO officially suports the NETGEAR N150 Wireless Adapter (WNA1100)
 with the native EV3 firmware. Of course, it also works with ev3dev.
-But ev3dev supports many other Wi-Fi adpters - if your Wi-Fi
+But ev3dev supports many other Wi-Fi adapters - if your Wi-Fi
 USB dongle works in your Linux system there is a good chance it
 will also works with ev3dev.
 
@@ -14,13 +14,15 @@ ev3dev. Even if your device is not in this list it may use the same
 chipset, so you should use lsub and dmesg commands to proper identify
 it after you connect it to your system:
 
-```
-~$ lsusb
-...
-Bus 001 Device 004: ID 050d:1102 Belkin Components F7D1102 N150/Surf Micro Wireless Adapter v1000 [Realtek RTL8188CUS]
-...
-```
+    ~^$ lsusb
+    ...
+    Bus 001 Device 004: ID 050d:1102 Belkin Components F7D1102 N150/Surf Micro Wireless Adapter v1000 [Realtek RTL8188CUS]
+    ...
 
+Adapters:
 * ThePiHut [Ralink RT5370, ID 148f:5370]
-* Belkin Components F7D1102 N150/Surf Micro Wireless Adapter v1000 [Realtek RTL8188CUS, ID 050d:1102]
+* Belkin Components F7D1102 N150/Surf Micro Wireless Adapter v1000 [Realtek RTL8188CUS, ID 050d:1102] (1)
+
+Notes:
+ 1 - The RTL8188CUS might have support problems in future distributions
 
--- a/docs/tutorials/supported-wi-fi-adapters
+++ b/docs/tutorials/supported-wi-fi-adapters
@@ -1,0 +1,26 @@
+---
+title: Supported Wi-Fi USB Adapters
+subject: Networking
+---
+
+LEGO officially suports the NETGEAR N150 Wireless Adapter (WNA1100)
+with the native EV3 firmware. Of course, it also works with ev3dev.
+But ev3dev supports many other Wi-Fi adpters - if your Wi-Fi
+USB dongle works in your Linux system there is a good chance it
+will also works with ev3dev.
+
+This is a list of other known Wi-Fi USB devices that also work with
+ev3dev. Even if your device is not in this list it may use the same
+chipset, so you should use lsub and dmesg commands to proper identify
+it after you connect it to your system:
+
+```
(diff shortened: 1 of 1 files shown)


dlech:
The website does not use GitHub flavor markdown, it uses [kramdown](http://kramdown.gettalong.org/) along with [liquid templating](https://github.com/shopify/liquid/wiki/Liquid-for-Designers).

So, for code samples, the code fence (```` ``` ````) does not work. Indent the lines 4 spaces instead.

We should also mention that although RTL8188CUS devices are very popular and seem to work well, upstream driver support for them is virtually non-existent, so they may not be a good choice in the long run. For example, [here is a recent discussion about the same drivers in Raspberry Pi](https://github.com/raspberrypi/linux/pull/1468).

dlech:
Also, this does not look like a tutorial - perhaps it belongs somewhere else?

JorgePe:
Sorry, I'll try to improve it.
It is no tutorial indeed but it seems the better place - is there any other place?
Also, is there any guideline and/or stylesheet to follow?
Perhaps I should discuss in another place, not in github?

WasabiFan:
> So, for code samples, the code fence ( ``` ) does not work. Indent the lines 4 spaces instead.

I am pretty sure that this is not the case. For example, [this page](http://www.ev3dev.org/docs/devtools/setting-up-the-ev3dev-build-ecosystem/#set-up-a-local-package-repository) uses both fences and indentation, and everything is being rendered correctly. We use it on a few other pages as well.

dlech:
> It is no tutorial indeed but it seems the better place - is there any other place?

There probably isn't right now. I think we could use another section under docs/ for general information. For example, we will need a place to put a table with a comparison of platform features (EV3, BrickPi, PiStorms, EVB) and other FAQ type stuff. What would make sense to you?


> Also, is there any guideline and/or stylesheet to follow?

Just have a look at the existing pages to see how they are done.


> Perhaps I should discuss in another place, not in github?

There is no better place than here.

dlech:
> I am pretty sure that this is not the case

Well, they have never seemed to work for me. :confused:

JorgePe:
At present, this is "Docs" structure:
```
Docs
 Getting Started
 Tutorials
 -
 Motors
 Sensors
 -
 More...
```

"More..." can have some sort of index to more documentation pages, including one for "Supported Devices" that could include a link to "Supported Wi-Fi Adapters".

By the way, isn't just "Motors" a bit misleading? The page correctly states "And everything else that plugs into output ports on the EV3". We have LEGO ports for I/O but we also have USB, Wi-Fi and BT...

WasabiFan:
> By the way, isn't just "Motors" a bit misleading?

I admit I didn't put much thought into the label text when I added that nav -- do you have a better word or phrase in mind? I haven't been particularly happy with the fact that most pages aren't obviously categorized by this structure, so I'm open to changing it.

---

> Well, they have never seemed to work for me. :confused: 

Hmmmm... It looks like there's a difference between local and GH Pages builds.

animation

JorgePe:
What about
```
Docs
 Getting Started
 Interfacing or I/O
 Tutorials
 More...
```
"Getting started" is important as THE entry point, specially for beginners.
Motors and sensors and actuators and even gadgets can fit in just one item. "Interfacing" doesn't sound good, "I/O" is probably better... This can link to an index page for the several types of interfaces EV3 can use, describing each before linking to particular devices
- Input ports -> sensors + advanced (I2C, UART, Analog...)
- Output ports -> motors and other actuators + advanced (I2C, UART...)
- native BT -> Networking
- USB -> Networking (Wi-Fi, Ethernet, BT, 1-wire, FTDI...) + tons of gadgets (webcam, TFT, NFC, RFID, smoke gen, relay board, sound... very difficult to categorize)

WasabiFan:
At some point, I'd like to move this discussion into a new issue to track the thing about nav.

For now, I'll just pile on and say that it is hilariously difficult to find the docs page for the tacho motor if you aren't intimately familiar with the site. This needs to be fixed.

dlech:
@WasabiFan and I were talking about this in his work on the Getting Started page. I think this deserves it's own page instead of being included in Getting Started or as a tutoral page as it stands now in this PR.

The page could either be `/wifi` or `/docs/wifi`.

We can have one section that is the "officially supported" adapters, namely the [ones that I own](https://github.com/ev3dev/ev3dev.github.io/pull/258#discussion_r81273026). It should also indicate that these adapters work with the official LEGO firmware.

Then another section can included "known working adapters" such as the ones already listed in the PR.

The info on using `lsusb` to get the device info is important too.

We can link to this page from Getting Started and from the Connecting with Wi-Fi tutorials.

WasabiFan:
Good idea; I agree that it should be on its own page. @JorgePe Would you like to update the page you have here to be what @dlech was describing, or should I merge this into my other docs work?

JorgePe:
sorry @WasabiFan didn't notice this question.

I'm not sure if most users will need this as a separate page - people with some linux experience will probably also have several wifi adapters already working.
Perhaps just a notice for newcomers is enough, with a link to some place in your docs?

WasabiFan:
For reference, the work that @dlech referred to is live at <http://www.ev3dev.org/docs/networking/>.

The separate vs. same page question comes down to, in my mind, the level of nuance that we want to explain. On one hand, having it inline where it's needed is useful because users don't have to get redirected out of their way to get this information. On the other hand, having a page where we can explain things in more detail may be helpful, and it's much easier to refer people to a separate page than to a section of a larger one.

I'm honestly not sure which I prefer at this point.

dlech:
There is a page for this on the wiki: https://github.com/ev3dev/ev3dev/wiki/USB-Wi-Fi-Dongles

PR #161, opened 2016-05-17
Write-up for how to set up remote debugging in PyCharm

eduvik:


rhempel:
I'm looking forward to trying this out - will review this week

fisherds:
Just adding my 2 cents (hoping other people will try this solution for Python development instead of letting this approach sit idol):
  I got this working (after a few failed attempts).  It's slick, but it was slow for me.  Not sure why, but it started fine then just got slower and slower.   I was using a WiFi connection.  I'll try to play with it more.  I like the idea here as the best Python development solution, but the slowness was killing me.  I admit it's a bit of a pain to setup so not for everyone, but it's got real potential as an alternative added to the end of @antonvh tutorial. http://www.ev3dev.org/docs/tutorials/setting-up-python-pycharm/ 

If this does get merged, I feel like the commit could use a few more details though.  Especially the info for setting up the deployment mapping.  I didn't know if I should use FTP, FTPS, SFTP, Local or mounted folder, or in place.  I tried Local or mounted folder first but that didn't seem to work.  Then I tried SFTP and it seemed to work.  Then I had to decide the root folder (turns out / worked best and put the /home/robot in the other area)

I think this path might indeed be the best development solution for .edu folks that can get a free copy of PyCharm.  It's very promising if I could get past my slowness issues.  I'm still playing with it, but wanted to share the thumbs up that this is valuable and important information for the tutorial for people that can get the professional version of PyCharm (free for .edu emails) and the slowness issue is not happening for everyone.

The other negative is that it gets a bit complex. :)  Maybe there is a better way.  Let me know if I did extra work.


### Remote interpreter setup
Screenshots of my remote interpreter setup:
remoteinterpreter

SSH Credentials for that remote interpreter:
sshcredentials

### Folder mapping to automatically transfer files
And my mapping
mapping

Full paths (that are obviously specific to me)
mappath

And my auto upload
upload


On all of those screenshots I was using my IP address just because I use a custom hostname instead of ev3dev.

Is there a way to do all that setup with a physical wire.  Would that be faster?  Is it the WiFi causing lag build up?  I'll have learn more about wired connections I guess.

WasabiFan:
Thanks for reviving this! Ralph's time estimates seem to be a bit... off :laughing: Not as bad as mine sometimes are though.

Would you be willing to open a new PR based off this one that includes the changes you wanted to make? I really haven't used PyCharm (and especially not its remote debugging features) but I definitely think that this is useful. This is one of the tutorials I've wanted to run through and update for a while now, but despite what I have convinced myself, I don't have much time to spend on these things :cry:

To open a new PR with these changes included, you should just be able to hit the "edit" button on GitHub from the branch with these changes. It will fork the repo for you and handle the branching; just make sure that when you open the PR you choose this repo as the base branch instead of @eduvik's.

fisherds:
Happy to help with a PR with more details.  I also want to play a bit more to see which part of the system was causing my slowness.  Was it...
1. Due to the connection method.  I was using WiFI does another approach reduce my slowness
2. Due to the automatic file transfer method.  I do like that the git repo on my computer is free to use a normal clone of a git repo on github btw.  However if that is the bottleneck I could just use the neat git trick from the tutorial and add a second remote to github.
3. Or the smoking gun, was my slowness due to the Remote execution.  It was neat to hit the play button from the PyCharm IDE, see the logs in the IDE, set breakpoints in the IDE etc, but it's not really that important.  I assume this one was the slowness culprit, but I want to investigate by setting up the Remote interpreter the same way again and changing other variables (like the connection method).

This whole approach is really only useful to people that can get the PyCharm professional version for free though.  
https://www.jetbrains.com/student/

@eduvik suggested that open source developers could get it free too???  I've only seen that it's free for .edu email JetBrain accounts.  The profession version costs $200 a year, so if you can't get it for free then I doubt many casual (non-edu) coders will care about this approach at all.

askvictor:
@fisherds - I don't know where to find the URL atm, but if you have a genuine oss project, you can apply to jetbrains for a free licence. 

I think there is also a speedup for remote debugging that can be installed on the remote device - the ide tells you how to do this whenever you start the remote debugger. I'm on mobile right now so can't find where it is, but that might help.

askvictor:
Here it is https://www.jetbrains.com/buy/opensource/

dlech:
IIRC, the free license is only available to a limited number of key contributors of an established open source project - it is not available for just anyone who says they are working on open source.

askvictor:
@dlech - the terms are on the link I provided.

WasabiFan:
I'm closing this in favor of #300, which includes these changes. Thanks @eduvik for getting things started!

PR #165, opened 2016-05-21
Add .nojekyll file when publishing using publish script

### LINKED ISSUE #162: Publish script should disable Jekyll
I believe there's a system for disabling Jekyll (something along the lines of adding a `.nojekyll` file). The publish script should do that so that GitHub doesn't try to build the static HTML.

WasabiFan:
Fixes #162.

I haven't published to confirm that this works end-to-end (because my Pages branch is currently preoccupied with another PR), but I know that this is successfully creating the right file and staging/committing it. Separately, I also timed the delay between push and web update on a different site that has a `.nojekyll` and it was less than 10 seconds. So overall, I expect that this will be an improvement.

### DIFF
--- a/publish.rb
+++ b/publish.rb
@@ -52,6 +52,9 @@ Dir.mktmpdir do |tmp|
 
     # having cname sends you annoying email
     FileUtils.rm 'CNAME'
+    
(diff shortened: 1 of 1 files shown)


dlech:
Looks good to me.

PR #168, opened 2016-06-04
Create controlling-wedo2-motor

JorgePe:
I found a way to use a linux system to control the new WeDO 2.0 motors.
Also found a python library that works with ev3dev.
Code examples aren't "clean" but I think it's a good example of how to use the EV3 to talk with BT 4.0 BLE devices.
The WeDO 2.0 hub is rather weak (just 2 AA batteries) but the same idea can be used with the SBrick to control Power Function motors or with those lots of IoT devices that keep appearing every day.
Of course, if someone has access to the WeDO 2.0 SDK, we could extend this tutorial to sensors and whatever we may find.


docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+
+The first WeDO version uses USB so every robot needs to be tethered to a «host»
+(usually a computer but can also be a Mindstorms EV3 running ev3dev)
+The second WeDO version uses BLE (Bluetooth Low Energy, a sub-set of the Bluetooth
+4.0 standard) so robots can now be totally autonomous.
+
+## Requirements
+
+The EV3 internal bluetooth is to old, just 2.1 compliant. To use BLE we need an
+USB Bluetooth 4.0 dongle supported by ev3dev. If it works with Ubuntu or with a
```
dlech: I have found that Bluetooth 4.0 != Low Energy. Technically, newer EV3s have a Bluetooth 4.0 chip, but not BLE support. BLE is an optional feature of Bluetooth 4.0.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+    	    TX bytes:1396 acl:0 sco:0 commands:32 errors:0
+	    Features: 0xff 0xff 0x2d 0xfe 0x9b 0xff 0x79 0x83
+	    Packet type: DM1 DM3 DM5 DH1 DH3 DH5 HV1 HV2 HV3 
+	    Link policy: RSWITCH HOLD SNIFF PARK 
+	    Link mode: SLAVE ACCEPT 
+	    Name: 'ev3dev'
+	    Class: 0x000000
+	    Service Classes: Unspecified
+	    Device Class: Miscellaneous, 
+	    HCI Version: 2.1 (0x4)  Revision: 0x0
```
dlech: As I said above, on newer EV3s, this will say `HCI Version: 4.0 (0x6)  Revision: 0x0` here, so we need to clarify this a bit.

Reference: <http://botbench.com/blog/2016/01/10/ev3-bricks-ble-or-blah/>

dlech:
FYI: We have a framework in place for user-defined sensors. This can be extended to create user-defined motors. So, in the future hopefully, we will have a program that bridges the bluetooth connection and creates motors in `/sys/class/dc-motor` to make using these easier.

dlech:
https://github.com/ev3dev/ev3dev/issues/521

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
+---
+title: Controlling a WeDo 2.0 motor
+subject: Networking
```
dlech: How about "Hardware - Motors" for the subject?

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+	    Service Classes: Unspecified
+	    Device Class: Miscellaneous, 
+	    HCI Version: 4.0 (0x6)  Revision: 0x1000
+	    LMP Version: 4.0 (0x6)  Subversion: 0x220e
+	    Manufacturer: Broadcom Corporation (15)
+
+In the above situation, `hci0` is our Bluetooth 4.0 BLE device (note "BUS: USB" and
+"HCI version: 4.0").
+
+We also need a recent bluez version, so please update your ev3dev to the latest
```
dlech: This step is probably not necessary. If there are any updates to the `bluez` package, they are probably just security updates.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+    gatttool -i hci0 -b A0:E6:F8:1E:58:57 --char-write -a 0x003d -n 01010100
+
+(You need to run this script with sudo, unless you already have root previleges)
+
+We see that it uses the gatttool command to send a sequence of 4 bytes to one specific
+handler (0x003d). The WeDO 2.0 has several handlers but until LEGO Education releases
+the promised SDK this is the only handler we «know» how to use:
+
+This is meaning of those 4 bytes:
+* the firt byte defines the port (01 or 02)
```
dlech: spelling firt/first

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+* the second byte defines the command (01 = motor speed)
+* the third byte defines the lenght of the following argument(s) (01)
```
dlech: spelling lenght/length

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+* the fouth byte is the argument, in this case the speed percentage
+
+To spin in one direction we send a positive value from 1 to 100 (or 01 to 64 in
+hexadecimal).
+To spin in the opposite direction we send a «negative» value from
+255 to 156 (or FF to 9C in hexadecimal).
```
dlech: why not write this as -1 to -100?

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+To stop the motor we set the speed as zero (00).
```
dlech: On WeDo 1.0, the value 128 (0x80) is used to brake the motor instead of coasting. Might be something to try out.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+## Python example
+
+To use pyhton with the WeDO 2.0 we need a BLE library. Unfortunately BLE
+support in python is still quite imature but at least pybluez offers an extension
+for BLE:
+
+    pip install pybluez
+    pip install pybluez[ble]
+
+This didn't work in my EV3 but since the ble extension is in fact gattlib I went for
```
dlech: It would be better to mention this before people install it instead of after if it is not likely to be useful.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+* Table of Contents
+{:toc}
+
+## Intro
+
+LEGO Education released the second version of WeDO in the beginning of 2016.
+We'll show how to use the bluez, the linux bluetooth stack, to wireless control a
+WeDo 2.0 motor.
+
+The first WeDO version uses USB so every robot needs to be tethered to a «host»
```
WasabiFan: Should these guillemets be quotes? For consistency, I don't think we need these characters (in a few other places as well).

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+it:
+
+    sudo apt-get install pkg-config libboost-python-dev libboost-thread-dev \
+     libbluetooth-dev libglib2.0-dev python-dev
+
+    pip install gattlib
+
+Please note that it takes A LOT of memory and around 2 hours time to install gattlib.
+After some failures ("virtual memory exhausted: Cannot allocate memory") I finally
+succeeded extending my ev3dev swapfile to 1024 MB.
```
dlech: Most people will not know how to do this. Can you show the commands?

e.g. https://github.com/ev3dev/ev3dev/issues/643#issuecomment-220512938

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,235 @@
...
+## A more practical example
+
+A BLE connection is not permanent - it drops after a few seconds. And the WeDO 2.0
+hub also enters in sleep mode a few seconds after the connection drops so we need
+to assure this never happens.
+
+We will use an EV3 touch sensor to control the direction of the WeDO 2.0 motor and
+periodically refresh the connection.
+
+    #!/usr/bin/python
```
dlech: If you want "pretty" code formatting, you can surround this with 

    {% highlight python %}
    ...
    {% endhighlight %}

and don't indent 4 spaces if you do this.

docs/tutorials/controlling-wedo2-motor.md
dlech: Gives it a bit of international flavor. :wink:

Another option would be to use markdown formatting to add emphasis here.

    *italic*
    **bold**

*italic*
**bold**

docs/tutorials/controlling-wedo2-motor.md
JorgePe: Yes, it might be better. I choosed Networking because most of the work is around bluetooth and also because I expect to extend the tutorial with sensor reading but I can make also an anothe tutorial just for for that.

docs/tutorials/controlling-wedo2-motor.md
JorgePe: I read somehwere (probably here, https://en.wikipedia.org/wiki/Bluetooth_low_energy) that BLE was included in the BT 4.0 standard but it is optional to implement "classic" or "both" modes - thats strange for a standard. I tought that since most BT 4.0 smartphones and BT 4.0 USB dongles support BLE then no one was opting for just "classic".
LEGO might have made a bad decision with the newer EV3 (and strange decision since it's using BLE with the WeDO 2.0) but as BT 4.2 is pushing for even more "Smart" / BLE features it's just a matter of time to have BLE in the EV3.

docs/tutorials/controlling-wedo2-motor.md
JorgePe: That's very interesting... that means newer EV3 might already support BLE. I ordered a kit a few months ago, need to pick it up at my LUG HQ to check that.

docs/tutorials/controlling-wedo2-motor.md
JorgePe: because I'm a complicated guy without enough programming experience to even consider that option? :)

docs/tutorials/controlling-wedo2-motor.md
JorgePe: i did iby editing "/etc/dphys-swapfile, just changing this line:
CONF_SWAPSIZE=1024

Then restarting dphys-swapfile:
/etc/init.d/dphys-swapfile stop
/etc/init.d/dphys-swapfile start

but since this is an ev3dev system that I've been updating almost since 2 years ago I'm not sure if dphys-swapfile is still used with latest image?

docs/tutorials/controlling-wedo2-motor.md
dlech: The newer versions don't support BLE. I've already tried it.

docs/tutorials/controlling-wedo2-motor.md
dlech: "Hardware - WeDo" would be good too.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -13,9 +13,9 @@ LEGO Education released the second version of WeDO in the beginning of 2016.
 We'll show how to use the bluez, the linux bluetooth stack, to wireless control a
 WeDo 2.0 motor.
 
-The first WeDO version uses USB so every robot needs to be tethered to a «host»
+The first WeDO version uses USB so every robot needs to be tethered to a *host*
 (usually a computer but can also be a Mindstorms EV3 running ev3dev)
-The second WeDO version uses BLE (Bluetooth Low Energy, a sub-set of the Bluetooth
+The second WeDO version uses [link](https://en.wikipedia.org/wiki/Bluetooth_low_energy) BLE (Bluetooth Low Energy, a sub-set of the Bluetooth
```
dlech: Why not make the link text "BLE"? It seems odd the way it is.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -133,37 +133,54 @@ Please note that for small speed values (less than 20%) the motor will not respo
...
+Unfortunately I couldn't make it work in my ev3dev system.
+
+Please note that it takes **a lot** of memory and around 2 hours to install gattlib.
 After some failures ("virtual memory exhausted: Cannot allocate memory") I finally
-succeeded extending my ev3dev swapfile to 1024 MB.
+succeeded extending my ev3dev swapfile to 1024 MB (please first check if you have
+enough free space in your SD card):
 
-This short python script makes the motor spin 2 second in each direction then stop:
+    sudo nano /etc/dphys-swapfile
```
dlech: You must be using an old image. We no longer use dphys swapfile.

JorgePe: Yes, I was afraid of that. I started all over again from a new SD card wit latest image to assure everything is correct.

docs/tutorials/controlling-wedo2-motor.md
JorgePe: OK.
So "subject" is dynamic, we can create our own subjects.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,255 @@
...
+    pip install pybluez[ble]
+
+Unfortunately I couldn't make it work in my ev3dev system.
+
+Please note that it takes **a lot** of memory and around 2 hours to install gattlib.
+After some failures ("virtual memory exhausted: Cannot allocate memory") I finally
+succeeded extending my ev3dev swapfile to almost 1 GB (please first check if you have
+enough free space in your SD card):
+
+    robot@ev3dev:~$ sudo dd if=/dev/zero of=/swapfile1 bs=1024 count=917504
```
dlech: `fallocate` is much faster than `dd` for creating files like this.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,255 @@
...
+    917504+0 records in
+    917504+0 records out
+    939524096 bytes (940 MB) copied, 442.332 s, 2.1 MB/s
+
+it will take 5 to 110 minutes to allocate space for such a big file
```
dlech: I think you mean 10 instead of 110, but it won't matter if you use `fallocate` instead, which takes very little time at all.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,255 @@
...
+    swapon: /swapfile1: insecure permissions 0644, 0600 suggested.
+
+this is a temporary measure so we'll skip security warning
+
+    robot@ev3dev:~$ sudo swapon -s
+    Filename				Type		Size	Used	Priority
+    /dev/zram0                             	partition	98300	8188	16383
+    /swapfile1                             	file    	917500	0	-1
+
+We don't want to use the swapfile in memory (it will overflow) so we disable it:
```
dlech: It will still overflow with both swap files enabled? I have never had to disable zram when adding an extra swap file. Maybe your experience was different?

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,255 @@
...
+
+    sudo swapoff /dev/zram0
+
+After installation please reboot your EV3 to reset the swapfile configuration then
+delete the big file as it is no longer needed.
```
dlech: You don't really need the reboot do you?

    sudo swapoff /swapfile1
    sudo rm /swapfile1

docs/tutorials/controlling-wedo2-motor.md
JorgePe: Yes, typo.
I considered fallocate but read that fallocate is not recommended in some file systems because the file might get "holes" (fragments?).

docs/tutorials/controlling-wedo2-motor.md
JorgePe: Yes, it overflowed several times at 10 minutes execution, always on same step. I think something tries to generate a large blob in the swapfile, bigger than the 96 MB that I had in zram.
After disabling zram it went straight way for more than a hour than failed because of permissions to copy package to some place, I re-run the install with sudo but need to see output this night.
"Top" shows sometimes more than 512 MB allocation, I'm not sure if  900 MB is really needed but 600 MB is not enough for sure.

docs/tutorials/controlling-wedo2-motor.md
JorgePe: Not sure - don't know if I can swapoff if the only swap file is in use (didn't try).

docs/tutorials/controlling-wedo2-motor.md
dlech: We are using `ext4` file system, so I don't think this is a problem.

docs/tutorials/controlling-wedo2-motor.md
dlech: I was going under the impression that you would not have to turn the zram swap off, but from what you said, it looks like you do have to do that.

You could turn the zram swap back on first here before turning the temporary swapfile off.

docs/tutorials/controlling-wedo2-motor.md
JorgePe: I'm not so sure, please see https://bugzilla.redhat.com/show_bug.cgi?id=1129205
Are there big differences between XFS and ext4 for this matter?

dlech: well, `fallocate` has seemed to work for me before, but `dd` is safer. I checked and `dphys-swapfile` uses `dd` as well. I'm thinking maybe we should make a new package that does all of this for you. We could make a systemd service that conflicts with the zram swap file so that starting the new service will automatically stop the zram swap. What you have here is good for now though.

By the way, you can use

    sudo systemctl stop zram_swap.service
    sudo systemctl start zram_swap.service

instead of 

    sudo swapoff /dev/zram0
    sudo swapon /dev/zram0

docs/tutorials/controlling-wedo2-motor.md
JorgePe: I confirm, it has to be "sudo pip install gattlib", as I was using root with the older image I wasn't sure.

docs/tutorials/controlling-wedo2-motor.md
```diff
@@ -0,0 +1,258 @@
+---
+title: Controlling a WeDo 2.0 motor
+subject: Hardware - WeDO
```
dlech: Typo: WeDO/WeDo

dlech:
merged. thanks!

JorgePe:
No need, I'm having fun. I wish I could help more, will keep trying at least some demos and weird niche usage scenarios like this.

PR #173, opened 2016-06-10
Remove plaintext dupes of names in project pages

WasabiFan:
I've noticed recently that many people are creating pages that include both their GitHub username and their plaintext name as separate authors. The text strings are pretty much just intended as a backup if someone puts arbitrary text in the "author" field; those cards look ugly, and in these cases the names are just duplicates of what already shows up as part of their profile.

Here I've tried to both clarify the template and update existing pages.

This only removes information from the page in one case, where the author of the PATHFIND3R project does not have their name on their profile.


dlech:
I assumed that these people want people to see their real name in addition to their github handle. The name attached to their github handle may or may not be their real name.

WasabiFan:
For 2 of the 3 authors, the text they supplied was exactly the same as that on their profile:

image image

The third one (as mentioned above) supplied a full name different from the name on their profile, so that one _does_ change with these modifications. However, my suspicion is that these authors were simply confused by the fact that our template included this:

```yaml
author: [ "@wasabifan", "The ev3dev Team" ]
```

I think they took that to mean that the user name went in the first slot and the textual name went in the second. That interpretation seems a lot more natural than the behavior we have actually set up.

But honestly, I don't think it matters whether they misunderstood or not; how 'bout we mention them (or at least the one whose page would change) in this thread and see what they want?

dlech:
> But honestly, I don't think it matters whether they misunderstood or not; how 'bout we mention them (or at least the one whose page would change) in this thread and see what they want?

good idea

WasabiFan:
@okanulas we have a question about your project page and were hoping you could weigh in. When you created it, you included two authors. One was your GitHub profile (`@okanulas`) and one was a full name (`Okan Ulas Gezeroglu`). With the way we have our site set up, each name you add is treated as a separate person. If you put in a GitHub username, the website displays the name on your GitHub profile. So, the "authors" section shows up like this:

image

Is this what you had intended? If you want your full name to appear, one way to do that is to set your name on your GitHub profile. If you would like to keep it as it is, that's fine too; we would just like to make sure that this is what you want the page to look like.

For reference, other pages look like this:

image

okanulas:
@WasabiFan I've updated my github settings to display my full name. It should be ok now. I referenced previous project pages when creating mine and didnt realized that each entry is a seperate person. I think that the 2nd entry at the authors template causes the misunderstanding. I'm going to remove the 2nd entry when i get home. There is only my mobile phone with me right now. Thanks for informing me.

WasabiFan:
> I referenced previous project pages when creating mine and didnt realized that each entry is a seperate person. 

Ah, sorry for the confusion! I was worried that we were confusing people with the way we had existing project pages set up, so your input is appreciated.

> I think that the 2nd entry at the authors template causes the misunderstanding.

We will change the template to better explain what the names are. Thanks for the feedback!

> I'm going to remove the 2nd entry when i get home. 

If you are OK with it, we can make that change for you so that you don't need to spend your time on it. That is what started this conversation; I proposed the removal of the extra entries, and we wanted to make sure that that was OK with the authors.

You can click on the "files changed" tab above to see the changes that I proposed. Is the modified template clearer, and is there anything else you think should be changed?

okanulas:
@WasabiFan Sorry for the delay. It is definely more clear that way. I think changing the 'author' tag to 'authors' makes it more clear that the each entry is a seperate person. 
`-authors: [ "@wasabifan", "The ev3dev Team" ]`

WasabiFan:
Oh, I didn't think of that... Good idea! I'll change that to be called "authors".

WasabiFan:
@dlech PR updated; can you take a look 👀?

dlech:
It's kind of weird to have `authors` on project pages but `author` on other pages. But, I guess it is OK.

Another option would be to change the sample to use a single author since that is the most common case and add a comment to show how to do multiple authors.

    author: "@dlech" # your GitHub user name
    # or make a list if there is more than one author
    # author: [ "@person1" "@person2" "person without GitHub account" ]

WasabiFan:
> It's kind of weird to have  authors  on project pages but  author  on other pages. But, I guess it is OK.

Yeah, I decided not to change the other pages because I really don't want to have to track down all the usages and test everything.

> Another option would be to change the sample to use a single author...

I think I'll do that instead (and just revert my previous commit). Keeps everything clean and simple.

WasabiFan:
:new:. Let's squash this when it's merged.

WasabiFan:
@dlech 🔔 ping; can you take a look at this?

dlech:
LGTM

PR #176, opened 2016-06-12
Let more people know about gitter

dlech:
Issue ev3dev/ev3dev#650

### DIFF
--- a/support.md
+++ b/support.md
@@ -97,6 +97,15 @@ Learn more about [writing on GitHub]{:target="_blank"}.
 </div>
 </div>
 
+Gitter
+------
+
+[gitter.im] is an online chat service that works in conjunction with GitHub.
+
+If you have a GitHub or Twitter account, come say hello at <https://gitter.im/ev3dev/chat>.
+
+[gitter.im]: https://gitter.im
+
 IRC
 ---
 


WasabiFan:
Could we merge the IRC and Gitter sections? There are some tips in the IRC section that are applicable to Gitter as well.

dlech:
Since Gitter has email notifications, you don't actually have to stick around. So really, "If no one ever answers you, do not be offended. The right person did not see your message, or maybe no one knows the answer. ~~Ask again later or open a GitHub issue.~~" is the only part that applies to both.

WasabiFan:
I was thinking about the part about people being in different time zones; That seems like something that would be helpful to others.

Also, I'm trying to get a Gitter badge for us on the ev3dev repo. Did you disable or never enable the section in the "Share this chat room" that lets you poke the @gitter-badger to give you a badge? Or maybe it's only available to admins? Despite my best efforts, I can't find the option that I'm pretty sure is supposed to be there. If you can't find that button, I can just manually open a PR.

dlech:
While looking for the gitter-badger thing just now, I saw that we can set up a welcome message for this room. I think this would be a better place to put the rules - where everyone will see them - instead of on the page in this pull request where only some people will see them.

dlech:
I think we don't have that option since the chat room is not tied to a single repository. But we're hackers, right? Like you said, we can just make our own badge.

[Join the chat at https://gitter.im/ev3dev/chat](https://gitter.im/ev3dev/chat?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)

WasabiFan:
Yep, I just wanted to experience the badger for myself :laughing: That badge should do it.

PR #177, opened 2016-06-12
Bring library info up-to-date in getting started and on libraries page

### LINKED ISSUE #175: Getting started guide should highlight libraries instead of drivers
At this point, I think that docs for the underlying drivers are less useful than links to libraries. So, I'd say that the libraries should be front-and-center in that section.

WasabiFan:
Fixes #175. It turns out that there were no links anywhere on the site to the `/docs/drivers` page, so this should also make that more discoverable.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -231,7 +231,7 @@ to connect for the first time. You can setup additional connections using the
 
 ## Step 6: Do something awesome
 
-The brick can run almost all languages that any other Linux distro can, so your
+The brick can run almost all programming languages that any other Linux distro can, so your
 favorite language is probably supported. Language bindings have already been
 written for many languages. **You can learn more about the available libraries
 [here](/docs/libraries).**
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -231,31 +231,14 @@ to connect for the first time. You can setup additional connections using the
 
 ## Step 6: Do something awesome
 
-This will be where we tell you how to use the EV3's main functions, and how
-to write programs. We are still learning, so everything here should be
-considered experimental and subject to major changes.
-
-### Using the EV3 hardware drivers
-
-Here are some guides for using each of the major components.
-
-* [Sensors](/docs/sensors)
-* [Motors](/docs/tutorials/tacho-motors/)
-* [Sound](https://github.com/ev3dev/ev3dev/wiki/Using-Sound)
-* [LCD](/docs/tutorials/using-ev3-lcd/)
-* [Buttons](/docs/tutorials/using-ev3-buttons/)
-* [LEDs](https://github.com/ev3dev/ev3dev/wiki/Using-the-LEDs)
-* [Bluetooth](https://github.com/ev3dev/ev3dev/wiki/Using-Bluetooth)
-
-### Writing programs
-
-There are a number of programming languages available to use. The brick can
-run almost all languages that any other Linux distro can, so your favorite
-language is probably supported. Language bindings have already been written
-for many languages. You can learn more about the available libraries [here](/docs/libraries).
+The brick can run almost all languages that any other Linux distro can, so your
+favorite language is probably supported. Language bindings have already been
+written for many languages. **You can learn more about the available libraries
+[here](/docs/libraries).**
 
 If the language you want isn't listed, you still can use it, but you'll have to
-do more of the heavy lifting yourself using the guides above. Once you get the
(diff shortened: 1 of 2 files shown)


docs/libraries.md
```diff
@@ -14,16 +14,15 @@ We have a repository of officially maintained language bindings, as well as many
...
-enhanced regularly.
+supports multiple implementations of the ev3dev API in a variety of languages.
 
 * Unified bindings:
     * [C++](https://github.com/ddemidov/ev3dev-lang-cpp)
     * [Node.js](https://github.com/wasabifan/ev3dev-lang-js)
     * [Python](https://github.com/rhempel/ev3dev-lang-python)
+    * Java
+      * [@mob41's ev3dev-lang-java](https://github.com/mob41/ev3dev-lang-java), which follows closely to the format of the other libraries.
+      * [@jabrena's ev3dev-lang-java](https://github.com/ev3dev-lang-java/ev3dev-lang-java), an interface more similar to that of LeJOS.
```
dlech: Does this library actually follow the ev3dev-lang spec? I know it started out that way, but I get the impression that it has changed to be LeJOS compatible instead, in which case it should be under the Extra Languages section.

WasabiFan: I put it there originally, but then moved it because I was annoyed that there was something called "ev3dev-lang" in the other section and was too lazy to deal with that :wink: I guess I will move it back down and just ignore the naming for now.

WasabiFan:
@dlech PR updated.

docs/libraries.md
```diff
@@ -40,14 +37,14 @@ for ev3dev is written using this library.
...
-
 ## Extra languages
 We also have many great contrubutors that are maintaining extra libraries for
 languages not included in our other repository.  Note that some of these
-libraries may be outdated due to the fast development cycle of ev3dev.
+libraries may be outdated due to the fast development pace of ev3dev.
 
 * Extra languages:
     * [Go](https://github.com/ldmberman/GoEV3) updated for ev3dev-jessie by @ldmberman, [original](https://github.com/mattrajca/GoEV3) by @mattrajca
+    * [ev3dev-lang-java](https://github.com/ev3dev-lang-java/ev3dev-lang-java) an interface similar to that of LeJOS by @jabrena
```
dlech: why put this between two Go libraries? wouldn't it be better to keep them together?

WasabiFan: Ugh, no idea why I did that.

WasabiFan:
:new:

docs/getting-started.md
```diff
@@ -231,31 +231,14 @@ to connect for the first time. You can setup additional connections using the
...
-* [LEDs](https://github.com/ev3dev/ev3dev/wiki/Using-the-LEDs)
-* [Bluetooth](https://github.com/ev3dev/ev3dev/wiki/Using-Bluetooth)
-
-### Writing programs
-
-There are a number of programming languages available to use. The brick can
-run almost all languages that any other Linux distro can, so your favorite
-language is probably supported. Language bindings have already been written
-for many languages. You can learn more about the available libraries [here](/docs/libraries).
+The brick can run almost all languages that any other Linux distro can, so your
```
dlech: might want to say "programming languages" in at least one place here so "languages" is not mistaken for i18n.

WasabiFan:
:new:

As a side note, I'm generally unhappy with how verbose our guides are; we have a lot of text where, for the most part, the link in the middle is really all that people care about. I can't figure out how to make that better without losing information that is important to some people, so I'm just leaving it as-is for now.

docs/libraries.md
```diff
@@ -14,16 +14,13 @@ We have a repository of officially maintained language bindings, as well as many
...
-[API specification](http://ev3dev-lang.readthedocs.org/en/latest/spec.html)
-that ensures that interface is almost identical for each, and they are being updated and
-enhanced regularly.
+supports multiple implementations of the ev3dev API in a variety of languages.
 
 * Unified bindings:
     * [C++](https://github.com/ddemidov/ev3dev-lang-cpp)
     * [Node.js](https://github.com/wasabifan/ev3dev-lang-js)
     * [Python](https://github.com/rhempel/ev3dev-lang-python)
+    * [@mob41's ev3dev-lang-java](https://github.com/mob41/ev3dev-lang-java)
```
dlech: This doesn't match the format of the other bindings listed here.

WasabiFan: Sorry, changed it in one place but not in the other. Fixed now.

dlech:
:shipit:

PR #170, opened 2016-06-08
Add tutorial for using Etcher

WasabiFan:
This adds an Etcher tutorial and updates related pages. The live preview is [here](http://wasabifan.github.io/ev3dev.github.io/docs/tutorials/writing-sd-card-image-etcher/), but there was an issue on the getting started page that broke one of the links and the site has yet to update since I fixed it.

Fixes ev3dev/ev3dev#660.

Note that the `<div>`s on the Etcher page that surround the screenshots were my way of forcing Kramdown to not wrap the screenshots in `<p>`s. If there's a better way to do that, let me know.

PR #178, opened 2016-06-15
Add universal API cache and implement proper fallbacks

WasabiFan:
This is a follow-up to #170.

This does three major things:
- Makes the API caching logic available generically instead of only for releases and uses that cache in user cards
- Adds fallback logic so that both cards and the downloads area will work if GitHub is unavailable
- Modifies card and download link logic so that they will be functional if JavaScript is disabled or broken (e.g. JQuery failed to load)

One thing I'd like to point out explicitly is [an addition to the getting started guide](https://github.com/ev3dev/ev3dev.github.io/compare/master...WasabiFan:add-universal-api-cache?expand=1#diff-a39c9f7aa728d5fa3b973bc6ba49228aR71). This content is shown by default and works as a fallback if the intelligent download buttons don't load properly (for the reasons mentioned earlier, among others). The first thing that happens when the document loads is that -- assuming JavaScipt and JQuery are both functional -- the deficient buttons are shown  and the fallback content is hidden. I say "deficient" because they don't actually have any `href` set on them by default. This means that the initial flash is minimal for users when everything is working properly. If, after attempting to load the buttons, it turns out that GitHub can't be reached, it hides the buttons and switches back to the fallback content.

I have made similar changes to the card logic so that cards will include fallback text in the static HTML until the JavaScript initializes them.

While I had originally disregarded the possibility of these issues, I realized that a large part of this site's audience is the type of person that might disable JavaScript to protect privacy. I figure this should fix any potential issues.

### DIFF
--- a/_includes/author-card.html
+++ b/_includes/author-card.html
@@ -10,7 +10,7 @@
 <!-- this looks weird, but it was the only way I could figure out how to test
      that the first character of include.author is "@" -->
 {% if at_sign.size > 0 and test_at_sign_is_first == 3 %}
-    <div data-card-user="{{author_no_at_sign}}" class="{{ extra_classes }}"></div><br/>
+    <div class="user-card {{ extra_classes }}" data-card-user="{{author_no_at_sign}}"><div class="user-card-name">{{ include.author }}</div></div></br>
 {% else %}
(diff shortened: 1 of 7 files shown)


dlech:
:shipit:

PR #182, opened 2016-07-12
Make release auto-detect logic accept bone-* images

WasabiFan:
This change should make the transition to image names that start with `bone-*` seamless; whenever the next round of images is released, the download buttons should be already prepared to serve the BeagleBone ones with the new name. We will still need to manually update the alternate text which appears when GitHub can't be reached.


dlech:
Awesome, thanks.

PR #183, opened 2016-07-15
Add "pro tip" to getting started recommending the "Use same image" batch option

WasabiFan:
Etcher has a button on the success screen that lets you run the operation again with the same image but a different card. This PR adds a panel with that information, as well as a final step in the "Flash" section that shows the success screen.

You can see the page [here](http://wasabifan.github.io/ev3dev.github.io/docs/getting-started/).


dlech:
:-1: There already too many alerts at this point and this is not likely to be useful to very may people. And most people can probably figure this out from reading the screenshot anyway.

WasabiFan:
So, I should remove the panel but keep the screenshot?

dlech:
Sure, if you want to.

PR #185, opened 2016-07-17
Remove link to a nonexistent "sensor tutorial"

WasabiFan:


dlech:
I guess if no one made on in the last 3 years, I shouldn't expect someone to make one in the next 3 years.

PR #186, opened 2016-07-24
Add C tutorial

WasabiFan:
This is a port of one of my other posts that worked well as a generic tutorial. In the future, it would be great to add reference pages for networking, file editing, and other basic things to make tutorials like this easier to follow. Also, this can probably be expanded to C++ fairly easily.

Some nice extra things that I would like to add in the future:
- Using multiple source files
- Linking external libraries

Preview here: http://wasabifan.github.io/ev3dev.github.io/docs/tutorials/getting-started-with-c/


WasabiFan:
> What I mean by this is that you give too many options

Although I didn't require use of a specific text editor/file transfer method, the toolchain, makefile, naming, etc. are well-defined and given in the tutorial. I could add instructions to require someone to install Filezilla, install a text editor, and then tell them what to do with each of those tools, but doing that in every "getting started with <language>" isn't sustainable and would fragment the important content. We should instead create separate pages that explain that process for beginners, hence the note about creating a networking and file editing tutorial.

dlech:
So in the tutorial, use the `nano` editor that ships with ev3dev. this way, there is no transferring of files. Keeps is as simple as possible.

dlech:
Or we could build a website like this on an RPi: http://www.delorie.com/djgpp/compile/ :smile:

WasabiFan:
:new:

I added a "Nano cheat sheet" page that I would expect to reference in every tutorial where we use nano. I would hope that we can expand on it and add other pages in the future.

dlech:
I kind of forgot about this PR. Do you think we still need it since we have http://www.ev3dev.org/docs/tutorials/using-docker-to-cross-compile/ now?

WasabiFan:
I have to say, the docker method was amazingly painless when I tried it out the other day (and on Windows, no less!). It is definitely the best option, and it's even easy enough for adventurous beginners to get started with. However, I see two reasons that I believe this page should still be added:

- This one is still easier for people who don't have any experience with the technologies involved. It explains everything in pretty extreme detail, where the other does not go in as deep; the docker one also is simply longer than this is, making it more imposing for new users. I do want to make a note here that there is a docker option, though.
- I see this tutorial as a good example of what a basic "getting started with _language_" tutorial should look like. It would be great to be able to reference it/refer other to it when making more involved tutorials, e.g. when eventually writing one for Python.

dlech:
Sounds good to me. :shipit:

rhempel:
Pre-built machines are the way to go - nice work putting this together @dlech !

PR #189, opened 2016-07-29
Update so it will work with old and new release

fryfrog:
Change instead of offering both options.


WasabiFan:
@dlech This seems fine to me; do you see any reason this shouldn't be merged?

dlech:
LGTM

WasabiFan:
@fryfrog Thanks for the pull request! It's always great to get fixes like this. Your change should show up on ev3dev.org within a few minutes. If you find any other issues in the docs/tutorials, don't hesitate to point them out or send a PR.

PR #192, opened 2016-08-03
Limit atom feed to 10 posts

### LINKED ISSUE #191: Limit item count in news rss/atom feed
I'm not sure what triggered it, but my rss reader decided that all of our news articles were new, which was 41. Most rss feeds are limited to the 10 most recent articles. This would also be nice for new subscribers.

WasabiFan:
Fixes #191.

### DIFF
--- a/news/atom.xml
+++ b/news/atom.xml
@@ -14,8 +14,11 @@ layout:
     <name>The ev3dev team</name>
   </author>
 
+  {% assign count = 0 %}
   {% for post in site.posts %}
   {% if post.categories contains "news" %}
+  {% assign count = count | plus: 1 %}
+  {% if count > 10 %}{% break %}{% endif %}
   <entry>
     <id>{{ site.github.url }}{{ post.id }}</id>
     <link type="text/html" rel="alternate" href="{{ site.github.url }}{{ post.url }}"/>


news/atom.xml
```diff
@@ -14,8 +14,11 @@ layout:
     <name>The ev3dev team</name>
   </author>
 
+  {% assign index = 0 %}
```
dlech: Why not use `forloop.index` instead of declaring your own index variable?

https://github.com/shopify/liquid/wiki/Liquid-for-Designers#helper-variables

news/atom.xml
```diff
@@ -14,8 +14,11 @@ layout:
...
   {% for post in site.posts %}
```
dlech: Or even better `{% for post in site.posts limit:10 %}

https://github.com/shopify/liquid/wiki/Liquid-for-Designers#helper-variables

news/atom.xml
WasabiFan: Because we need to filter the posts to get only news items. We want ten news items, not all the news items in the first ten posts. There are features in Jekyll 3.2 that would make this work ([the `where_exp` filter](https://github.com/jekyll/jekyll/pull/4478)), but GH pages is still on 3.1.

This applies to the second comment as well.

dlech: I missed that part. Maybe better to call it `count` instead of `index` in that case.

WasabiFan: Yeah, I missed it too until I got into the code. I've amended my commit to rename that variable.

dlech:
LGTM

PR #195, opened 2016-08-05
Add docker cross-compiler info

dlech:


WasabiFan:
Just curious: why do you specify PowerShell instead of CMD? I've used both with Docker and there aren't any differences as far as I can tell, aside from the fact that pasting is a different menu.

dlech:
Because I don't know what I'm doing. :wink:

WasabiFan:
Oh, ok :laughing:. It isn't a problem -- I was just wondering if there was something I had missed.

WasabiFan:
> We think you will have a whale of a good time.

😆 :+1:

dlech:
I had to make sure to include the picture so that people will get the joke. :smile:

PR #196, opened 2016-08-05
Minor typo fixes and clarification in docker tutorial

WasabiFan:
I noticed a few typos and clarification opportunities while reading that I figured I'd fix. I tried to keep line wrapping at the original length, so some lines show changes that aren't really changes.


dlech:
Thanks. (I've been chatting with someone on irc about this, so I published without waiting for a proper review.)

PR #197, opened 2016-08-09
x3d support

bmegli:
The x3d support that I need for Mapping Project page. 

I want to have this kind of mapping results:
http://htmlpreview.github.io/?https://github.com/bmegli/ev3dev-mapping-results/blob/master/point%20clouds/html/crime%20scene/crime_scene.html

Embeded on ev3dev projects page (I already have them on my local github.io)

Now, this pull consists off:
-x3d css link in head includes
-x3d scripts link in head includes
-new include for x3d-model

While this is working ok I am not sure we need that in head of all ev3dev pages. 

I would happilly just throw it to project page but this is auto-genarated with Jekyll with html head already in place. Unfortunately I need css link specifically in html head.

So maybe I am worrying too much, a css link and script is not a problem.
But maybe somebody has better idea how to make exception for one project post.


WasabiFan:
In the front matter of your page (between the `---`s), you can set variables. You can then access them in the head section to optionally include the scripts. So, something like this should work to optionally include those scripts:

In your page:
```markdown
---
include_x3d_scripts: "1"
---

Page content
```

In the `head` template:
```handlebars
{% if page.include_x3d_scripts %}
<script ...>
{% endif %}
```

_includes/x3d-model.html
```diff
@@ -0,0 +1,10 @@
+{% comment %}
+
+Used to format x3d model
+
+Parameters:
+
+source: The url of the x3d model
+caption: (optional) A caption that is displayed under the image.
+
+{% endcomment %}<span class="screenshot"><x3d><scene><inline url="{{ include.source }}"></inline></scene></x3d> {% if include.caption %}<small>{{ include.caption }}</small>{% endif %}</span>
```
WasabiFan: This is invalid HTML, right? `x3d` and `scene` aren't HTML tags? This isn't a big problem if that is what this library requires, but overall it seems like a bad practice on their part...

bmegli:
@WasabiFan Thanks for suggestions!

Yes, I definitely am not an expert on web development.

Any suggestions with 3d models includes?

I can simply not use template for 3d content and use the x3d tags only in project page where I know they have the right css and scripts (through include_x3d_scripts: "1")

_includes/head.html
```diff
@@ -16,6 +16,10 @@
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
     <link rel="stylesheet" href="/stylesheets/bootstrap.css">
     
+	<!-- X3D styles and scripts -->
```
dlech: we are using 4 spaces for indent, not tabs

_includes/head.html
```diff
@@ -16,6 +16,10 @@
...
+	<link rel='stylesheet' type='text/css' href='http://www.x3dom.org/download/x3dom.css'></link>
```
dlech: Apparently `</link>` is not needed and is causing errors in the html checker.

dlech:
> include_x3d_scripts: "1"

I would rather see this written `include_x3d_scripts: true`

bmegli:
@dlech, yes, I will follow that

@WasabiFan 
> This is invalid HTML, right? x3d and scene aren't HTML tags? This isn't a big problem if that is what this library requires, but overall it seems like a bad practice on their part

Oh, ok. You mean that.

Yes, x3d is not part of html5, it's a separate ISO standard. But it seems currently  widely supported 3D web standard and x3dom works on top of WebGl in the worst case, so it should be well supported. 

At least 4 major browsers support WebGL natively (Firefox, Chrome, IE, Safari) and lots of other support WebGL too.

http://www.x3dom.org/contact/

X3D is Web3D standard and X3DOM is somewhat like official way to do it on Web3D pages.

You guys are such a purists!

I like it ;-)

I don't think it's a problem functionally. In the worst case, if the browser doesn't support it I can display some text or if I have time and patience to do it, replace it by screenshot as fallback.

_includes/head.html
WasabiFan: I think it needs to be self-closing becuse links can't have internal content.

WasabiFan:
> I can simply not use template for 3d content and use the x3d tags only in project page where I know they have the right css and scripts (through include_x3d_scripts: "1")

I think the template is fine, you just have to make sure that you include the scripts everywhere you use the template. It's not optimal, but it works.

> I would rather see this written include_x3d_scripts: true

This is one of those cases where I know I have tested that before but can't remember why I avoided it :laughing: If it works, feel free to use it, but make sure that you're confident it is doing what it should be.

_includes/head.html
bmegli: For reference:

It's not allowed in html5 for link: neither </link> and not even <link some content/>
It's mandatory in xhtml, so in xhtml <link some content />

So for pure html: it should be <link ...>
For xhtml <link .../>

The browsers ignore / the at the <link/>.

I will follow pure html5 then.

bmegli:
Ok, I have added:
- conditional inclussion based on enable_x3d_support variable
- if/else to x3d-model include that instructs how to enable x3d support if someone uses it without enable_x3d_support: true in YAML front matter
- fixed link closing tag problem
- added some comments
- converted tabs to 4 spaces

dlech:
Are you going to add the change to the project page so we can actually if this works as expected?

bmegli:
Yes. I am "just" finishing the project page so this can wait and go together.

I am evaluating the ways to compress x3d models. 4 MB is just too much and that's mainly because of xml representation.

Hopefully today.

bmegli:
Ok, so I am closing this one and it goes along with the project page.

PR #200, opened 2016-08-10
Add ability to override project thumbnails with static image

WasabiFan:
This also fixes an issue I noticed with wrapping when in the "small" display size range.

Can hopefully be used in #198.

Issue #199, opened 2016-08-10
Use tiling library instead of Bootstrap grid on index pages


WasabiFan:
The "projects" page is having issues with scaling and flow because we are using hard-coded rows and Bootstrap classes. We should be using one of the existing tiling libraries that works with Bootstrap.

PR #202, opened 2016-08-11
Use Masonry for project page tiling (plus some other stuff 'cause I am lazy)

WasabiFan:
This is a combination of #200 and a fix for #199. The two commits correspond to two sets of changes.

The project page can be seen live [here](http://wasabifan.github.io/ev3dev.github.io/projects/).

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -34,6 +34,10 @@
     <script src="/javascripts/jquery.loadTemplate-1.4.4.min.js"></script>
     <script src="/javascripts/respond.js"></script>
     <script src="/javascripts/ua-parser.min.js"></script>
+    <script src="https://npmcdn.com/imagesloaded@4.1/imagesloaded.pkgd.min.js"></script>
+    {% if page.include_masonry %}
+    <script src="https://npmcdn.com/masonry-layout@4.1/dist/masonry.pkgd.min.js"></script>
+    {% endif %}
 
     <!-- Caches data from requests to API endpoints -->
     <script src="/javascripts/api-cache.js"></script>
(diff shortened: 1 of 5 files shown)


WasabiFan:
I don't think we should squash when this is merged because of the separate logical changes.

_includes/head.html
```diff
@@ -34,6 +34,10 @@
     <script src="/javascripts/jquery.loadTemplate-1.4.4.min.js"></script>
     <script src="/javascripts/respond.js"></script>
     <script src="/javascripts/ua-parser.min.js"></script>
+    <script src="https://npmcdn.com/imagesloaded@4.1/imagesloaded.pkgd.min.js"></script>
```
dlech: should this also be inside the `if page.include_masonry` or does it do other things?

dlech:
It's not looking so good in Firefox 48 (Linux)

selection_009

dlech:
Um.. that's not the screenshot...

dlech:
Trying again...

It's not looking so good in Firefox 48 (Linux) when you clear your browser cache and reload.

selection_009


Looks great if you reload a second time though.

_includes/head.html
WasabiFan: This helper is called even if the jQuery set it is given is empty with my current usage. I did it that way because a) this script is useful for other things and b) I don't want to have to manage/load another 3-line script file for one page. It's this code that uses it:

```js
$('#project-thumbnail-grid').imagesLoaded().progress(function() {
    //...
}
```

WasabiFan:
> Looks great if you reload a second time though without clearing the cache.

That makes me think that the problem is that the code that updates layout on image load isn't working.

I just made a change (amended the commit). Can you try now?

dlech:
It working correctly now. :shipit:

PR #216, opened 2016-08-22
Accommodate taller images in screenshots

WasabiFan:
Images of a taller aspect ratio were prevented from being full-width because of this rule, so this loosens it up a bit.

### DIFF
--- a/stylesheets/site-structure.scss
+++ b/stylesheets/site-structure.scss
@@ -90,7 +90,7 @@ h1.panel-title {
 }
 
 .screenshot > a > img {
-    max-height: 480px;
+    max-height: 640px;
 }
 
 .inline-screenshot {


stylesheets/site-structure.scss
```diff
@@ -90,7 +90,7 @@ h1.panel-title {
 }
 
 .screenshot > a > img {
-    max-height: 480px;
+    max-height: 640px;
```
dlech: I don't like scrolling so much, which IIRC is why this was set to 480px. If an image is bigger, you can click on it to view the full size image.

WasabiFan: No, the problem wasn't that the image was "to small" _per se_; the issue was that this rule prevented images from being full-width on the page. That made it look extremely awkward when full-width text was before and after an image that wasn't full-width.

The project page that originally made me notice the issue isn't using the screenshot template anymore, so even if other pages are affected I'm, not going to bother with fixing it at this point.

PR #218, opened 2016-08-27
Create upgrading-ev3dev-using-shell

rikblok:
I couldn't find all this information in one place and thought it might be useful.


docs/tutorials/upgrading-ev3dev-using-shell
```diff
@@ -0,0 +1,24 @@
+---
+title: Upgrading ev3dev Using Shell
+subject: Upgrading
+---
+
+Once you've written the ev3dev image to the SD card, booted, and setup a network 
+connection, you may want to update to the latest kernel.  To do so:
+
+1. Connect to the EV3 using SSH.  (Wireless or USB.)  The default login is
```
WasabiFan: ev3dev supports hardware other than EV3s (e.g. BrickPi, PiStorms, EVB) so maybe call it "the ev3dev device."

docs/tutorials/upgrading-ev3dev-using-shell
```diff
@@ -0,0 +1,24 @@
+---
+title: Upgrading ev3dev Using Shell
+subject: Upgrading
```
WasabiFan: I'm not sure that "upgrading" is a broad enough topic name. Maybe "core" or "system"?

docs/tutorials/upgrading-ev3dev-using-shell
```diff
@@ -0,0 +1,24 @@
...
+---
+
+Once you've written the ev3dev image to the SD card, booted, and setup a network 
+connection, you may want to update to the latest kernel.  To do so:
```
WasabiFan: It would be nice to make a note here that you are also upgrading the drivers, so there may be changes that break user software. Maybe include a link to the news feed (`/news`, I think) and say that one should make sure that they are aware of breaking changes before upgrading.

docs/tutorials/upgrading-ev3dev-using-shell
```diff
@@ -0,0 +1,24 @@
+---
+title: Upgrading ev3dev Using Shell
+subject: Upgrading
+---
+
```
WasabiFan: Upgrading can take a long time; sometimes even hours. A warning about this would be nice. You can add formatting to put text in a "warning" alert like this:

```
{% include icon.html type="warning" %}
Alert text goes here
{: .alert .alert-warning}
```

docs/tutorials/upgrading-ev3dev-using-shell
```diff
@@ -0,0 +1,24 @@
...
+username=robot, password=maker.
+
+2. Execute these commands in sequence:
+
+    sudo apt-get update
+    sudo apt-get upgrade
+    sudo apt-get dist-upgrade
+    sudo reboot
+
+3. Wait for each command to complete before typing the next.  Some steps require
```
WasabiFan: It might make more sense for this first sentence (the one about doing them one at a time) to be in the previous step to go with the commands.

docs/tutorials/upgrading-ev3dev-using-shell
```diff
@@ -0,0 +1,24 @@
...
+    sudo apt-get update
+    sudo apt-get upgrade
+    sudo apt-get dist-upgrade
+    sudo reboot
+
+3. Wait for each command to complete before typing the next.  Some steps require 
+confirmation to proceed.  Press 'y' when asked.  The `upgrade` step may take up 
+to a couple hours to complete.
+
+4. After the brick has rebooted, press `About` on the brick's menu to confirm it
```
WasabiFan: Good thought! I like the idea of having people look at Brickman to check the kernel version.

WasabiFan:
Thanks, I definitely think we need this!

I added some comments; they're just cosmetic/style changes for the most part, the content itself is great.

WasabiFan:
I just realized that the file is missing an extension (which is required for it to render properly). Can you add `.md` to the end?

WasabiFan:
There are some formatting issues right now; the code isn't formatted as code, and the code and text/alert aren't inside the list element. The problem is... I can't figure out how to fix it. @dlech do you know?

WasabiFan:
Other than the formatting, this looks good to me.

docs/tutorials/upgrading-ev3dev-using-shell.md
```diff
@@ -0,0 +1,32 @@
...
+
+Wait for each command to complete before typing the next.  Some steps require 
+confirmation to proceed.  Press 'y' when asked.  
+
+{% include icon.html type="warning" %}
+The `upgrade` step may take a couple hours to complete.
+{: .alert .alert-warning}
+
+3. After the brick has rebooted, press `About` on the brick's menu to confirm it
+has a current kernel.
```
dlech: screen shot would be nice here.

dlech:
This is a good start, but I think it could be much more informative. For example...

* Explain that `apt-get update` gets a list of packages from the package repositories. If you don't do this, subsequent commands will most likely fail.
* Explain that `apt-get upgrade` upgrades *all* packages. You rarely want to do this unless you like sitting around for a long time waiting for it to complete or you are the kind of person who has to have everything up to date.
* You can upgrade individual packages with `apt-get install <package-name>`. This is a better option for most people.
* To upgrade the kernel, run `apt-get install linux-image-ev3dev-ev3` (or `-rpi` or `-rpi2` or `-bb.org` if you have RPi 0/1, RPi 2/3 or BeagleBone respectivly)
* You only need to reboot if the kernel was actually upgraded.

dlech:
As for the list formatting, the blocks that are new paragraphs of the same list item need to be indented as much as the text of the first item. I like to put two spaces between `1.` and the text so that additional blocks are indented 4 spaces as is common in markdown. Also, you will probably have to use html tags instead of liquid templating within list items to make it work.

```
1.  Item 1
2.  Item 2

    Continuing item 2

    <div class="fancy">Fancy thing in item 2</div>

3.  Item 3
```

WasabiFan:
> This is a good start, but I think it could be much more informative. For example...

Good ideas.

> As for the list formatting, the blocks that are new paragraphs of the same list item need to be indented as much as the text of the first item... Also, you will probably have to use html tags instead of liquid templating within list items to make it work.

I couldn't get that to work. I added `<pre>` and `<code>` tags around the code, but the area of the `<pre>` was 2x taller than necessary. I suspect it's an issue with CSS though.

dlech:
If you have code within a list item, indent it 8 spaces instead of 4

WasabiFan:
@rikblok Are you able to make the additions that @dlech suggested? This page is great and we really do need upgrade instructions, so I would like to get this finished. If you don't have the time to do it, that's totally fine; I can finish it up for you if you'd like.

rikblok:
Thanks @WasabiFan, I'm going to busy for the next few weeks so it would be great if you can finish it up.

Thanks for all your help!

WasabiFan:
No problem; I'll branch off of your changes when I can and make that fix. Thanks for contributing!

WasabiFan:
I've built upon the changes here and opened #236 as a result.

PR #221, opened 2016-09-03
Rewrite and restructure libraries page

WasabiFan:
This makes it less about politics and implementation and more about the things that a user needs to know.

I sorted the languages in order of ease-of-use; essentially, they are in the order a beginner should see them in. I figure that if an expert is reading the page, the order won't matter, because they will find a specific section directly anyways. I wasn't really sure where to put ev3devKit, as languages like Vala are easy to use but setting up a compiler is more difficult, in addition to the fact that -- due to the nature of it -- there aren't any language-specific instructions.

~~Note that I decided to put <https://github.com/mob41/ev3dev-lang-java> in the "not up-to-date" section because it does not support recent drivers, but there has been activity in the last few days from the author saying that they have been busy but plan to get it updated. That library might need to be moved to the other section if/when that happens.~~ ~~Edit: Not only were changes just pushed to that repo (2 minutes ago as of this edit), but the kernel that was most recently supported was a version that included all significant breaking changes so far. So, I think I will need to move this into the other section and add a description.~~ Edit 2: Nevermind. https://github.com/mob41/ev3dev-lang-java/commit/bdb5b8924629d4cba1c59ada50a8c67d7e26da7c.

Preview here: http://wasabifan.github.io/ev3dev.github.io/docs/libraries/.

ev3dev/ev3dev#733

### DIFF
--- a/docs/libraries.md
+++ b/docs/libraries.md
@@ -4,32 +4,42 @@ subtitle: Control ev3dev devices from code
 excerpt: "If you are looking to write a program that takes advantage of motors, sensors, or other native devices, using a language binding is the way to go. These are the best options for each language."
 ---
 
-Before you can start writing code that runs on ev3dev, you need to choose a programming language. Community members have built libraries that let you use features
-such as motors and sensors from your code. Choose the language you want from below and follow the link to the library's homepage get started.
+Before you can start writing code that runs on ev3dev, you need to choose a
+programming language. Community members have built libraries that let you use
+featuressuch as motors and sensors from your code. Choose the language you want
+from below and follow the link to the library's homepage get started.
 
 ## Python
 <https://github.com/rhempel/ev3dev-lang-python>
 
-Python is a high-level, general-purpose, interpreted scripting language. It is great for beginners, while also powerful for experienced coders.
-If you are new to programming and unsure which language to choose, Python is a great choice!
+Python is a high-level, general-purpose, interpreted scripting language. It is
+great for beginners, while also powerful for experienced coders.If you are new
+to programming and unsure which language to choose, Python is a great choice!
 
-**Warning!** There is another Python library named `python-ev3` created by @topikachu which is _not_ the same as this one. When searching for documentation online, make sure that you are reading about @rhempel's `ev3dev-lang-python`.
+**Warning!** There is another Python library named `python-ev3` created by
+@topikachu which is _not_ the same as this one. When searching for documentation
+online, make sure that you are reading about @rhempel's `ev3dev-lang-python`.
 
 [Get started with Python](https://github.com/rhempel/ev3dev-lang-python){: .btn .btn-default }
 
 ## JavaScript
 <https://github.com/wasabifan/ev3dev-lang-js>
 
-JavaScript is the scripting language of the Web. With a program called Node.js, you can write JavaScript that can be run locally like any other scripting language.
-JavaScript with Node.js is great for writing web servers and other asynchronous programs which don't need to run tight loops or precise timers. It has a syntax similar to that of C and related
-languages, and is relatively easy to learn.
+JavaScript is the scripting language of the Web. With a program called Node.js,
+you can write JavaScript that can be run locally like any other scripting
+language. JavaScript with Node.js is great for writing web servers and other
+asynchronous programs which don't need to run tight loops or precise timers. It
+has a syntax similar to that of C and related languages, and is relatively easy
+to learn.
 
 [Get started with JavaScript](https://github.com/wasabifan/ev3dev-lang-js){: .btn .btn-default }
 
 ## Go
 <https://github.com/ev3go/ev3dev>
 
-Go is a compiled, statically-typed language created at Google. It aims to be simple and light while still providing modern language features. While it is a compiled language, cross-compilation is _not_ required to run code on ev3dev,
+Go is a compiled, statically-typed language created at Google. It aims to be
+simple and light while still providing modern language features. While it is a
+compiled language, cross-compilation is _not_ required to run code on ev3dev,
 unlike other compiled languages; this makes Go much simpler to work with.
 
 [Get started with Go](https://github.com/ev3go/ev3dev){: .btn .btn-default }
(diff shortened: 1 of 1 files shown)


WasabiFan:
@ndward Can you take a look at the revised page (the preview is [here](http://wasabifan.github.io/ev3dev.github.io/docs/libraries/)) and give us your thoughts on it? Specifically, does it make it easier to use in general and does it address your concerns? Is there anything that you think should be further improved?

Note that so far this only changes the "Libraries" page. Other pages haven't been updated yet, but this seemed like a good first step.

docs/libraries.md
```diff
@@ -1,52 +1,83 @@
...
+JavaScript is the scripting language of the Web. With a program called Node.js, you can write JavaScript that can be run locally like any other scripting language.
+JavaScript with Node.js is great for writing web servers and other asynchronous programs which don't need to run tight loops or precise timers. It has a syntax similar to that of C and related
+languages, and is relatively easy to learn.
+
+[Get started with JavaScript](https://github.com/wasabifan/ev3dev-lang-js){: .btn .btn-default }
+
+## Go
+<https://github.com/ev3go/ev3dev>
+
+Go is a compiled, statically-typed language created by at Google. It aims to be simple and light while still providing modern language features.
```
dlech: typo "created by at"

one preposition will do :smiley:

dlech:
There are some other unmaintained libraries that have been dropped from our list, such as R and some unfinished ones such as C#. It would be nice to add these to the list so that if there is new interested, at least a person will have a starting place instead of starting from scratch.

docs/libraries.md
```diff
@@ -1,52 +1,83 @@
 ---
-title: Libraries
-subtitle: Pre-made language bindings that make accessing the EV3 device drivers easy
-excerpt: "If you are looking to write a program that takes advantage of the EV3's motors, sensors, or other native devices, using a language binding is the way to go."
+title: Interface libraries
```
dlech: I think I like the title as the more simple "Libraries"

The subtitle can make it clear that these libraries are for interfacing with hardware.

docs/libraries.md
```diff
@@ -1,52 +1,83 @@
...
+## Python
+<https://github.com/rhempel/ev3dev-lang-python>
 
-## Unified Libraries
+Python is a high-level, general-purpose, interpreted scripting language. It is great for beginners, while also powerful for experienced coders.
+If you are new to programming and unsure which language to choose, Python is a great choice!
 
-Our official "unified" [language binding repository](http://github.com/ev3dev/ev3dev-lang)
-supports multiple implementations of the ev3dev API in a variety of languages.
+**Warning!** There is another Python library created by @topikachu which is _not_ the same as this one. When searching for documentation online, make sure that you are reading about @rhempel's `ev3dev-lang-python`.
```
dlech: It would be good to add the name of the other library as well, e.g.

> There is another Python library **named `python-ev3`** created by...

docs/libraries.md
dlech: Another awesome advantage of go is that it has a built-in cross compiler, so even though this is a compiled language, you don't have to mess with setting up our Docker image or another cross compiler to get code that will run on the EV3.

dlech:
Lua is another out of date language that deserves a link.

dlech:
I think there should also be a section for up-to-date alternatives. For example ev3dev-c and ev3devKit both have python bindings. ev3dev-c also has ruby and perl bindings. ev3devKit works in C (although I don't recommend it).

there are probably some other libraries that people haven't told us about either, such as https://github.com/noanoa07/ev3dev_ruby

WasabiFan:
> I think there should also be a section for up-to-date alternatives. For example ev3dev-c and ev3devKit both have python bindings. ev3dev-c also has ruby and perl bindings. ev3devKit works in C (although I don't recommend it).

What do you mean? AFAIK, the Python one is up-to date for the most part, so they don't need alternatives... or am I missing something?

---

Good points. I'll add a note about contributing to the out-of-date libraries and add the ones that you mentioned (as well as any others I find) later today.

docs/libraries.md
WasabiFan: Yeah, I was trying to come up with a more descriptive name, but I don't think this is any better.

docs/libraries.md
WasabiFan: I don't have any experience with Go, so I was really just winging it; thanks for pointing that out.

dlech:
I mean that people looking for a python library won't know that ev3dev-c is an alternative to ev3dev-lang-python unless we point out that ev3dev-c also has python bindings in addition to being a C library.

WasabiFan:
Ah, that makes sense. I'll add a note about it.

WasabiFan:
Oh! You meant "alternatives that are up to date as well" and not "alternatives that would let you use that language with an up-to-date library". I thought you were saying that we needed to suggest alternatives because the existing Python one shouldn't be used.

WasabiFan:
Changes made. @dlech can you take another look?

docs/libraries.md
```diff
@@ -1,52 +1,144 @@
...
+to learn.
+
+[Get started with JavaScript](https://github.com/wasabifan/ev3dev-lang-js){: .btn .btn-default }
+
+## Go
+<https://github.com/ev3go/ev3dev>
+
+Go is a compiled, statically-typed language created at Google. It aims to be
+simple and light while still providing modern language features. While it is a
+compiled language, cross-compilation is _not_ required to run code on ev3dev,
```
dlech: Cross-compilation is required. A separate cross-compiler is not - it is built in.

docs/libraries.md
```diff
@@ -1,52 +1,144 @@
...
+<div class="panel-body">
+<p>
+    Use them with caution, as some functionality will likely be broken. If you
+    see a library below that you'd like to see in a better state, consider
+    contributing to it to get it updated and ready to use.
+</p>
+</div>
+</div>
+
+## C#
```
dlech: The `#` in `C#` is not showing up on the preview site. You might need to escape it (`C\#`)

dlech:
Aside from the couple technical issues, I think this looks good.

docs/libraries.md
WasabiFan: 😐

docs/libraries.md
WasabiFan: Oh, that makes sense... good that you noticed!

WasabiFan:
I'll take that to mean :shipit: once the CI server is happy

PR #228, opened 2016-09-05
Add SSH tutorial and remove mentions of empty page

WasabiFan:
Fixes ev3dev/ev3dev#727

Preview at http://wasabifan.github.io/ev3dev.github.io/docs/tutorials/connecting-to-ev3dev-with-ssh/.

Also, can we remove the "these instructions are for..." version warning message? The December image has been in the wild for almost ten months at this point, so I doubt that anyone is using an image from before that. It isn't a big deal, but it would be nice to remove an alert from that page if it isn't needed anymore.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -240,6 +240,10 @@ to connect for the first time. You can setup additional connections using the
 `connmanctl` command.
 {: class="alert alert-info"}
 
+Now that you have a network connection, you should
+[connect to your EV3 with SSH](/docs/tutorials/connecting-to-ev3dev-with-ssh)
+if you haven't done so already.
+
 </div>
 </div>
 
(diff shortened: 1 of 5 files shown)


dlech:
I think it would be good to have a link from the network connection tutorials to the ssh tutorials. There is not much else you can do with the network connection out of the box, so it is the logical next step. Also, we could use a link to the ssh tutorial from the getting started page the first time we show a command that is typed into an ssh terminal.

About the version warnings, I'll want to put them back next time we make a breaking change (like ev3dev/ev3dev#721), so better just leave them for now.

WasabiFan:
I left out the links because I expect a lot of the link paths to change when I do a major re-evaluation of documentation flow. Now that I think about it, I can probably add those links that you mentioned under the assumption that they will still be needed, or at least will be helpful in the interim.

WasabiFan:
:new:

dlech:
LGTM

PR #231, opened 2016-09-05
Update readme for ubuntu 16.04

dlech:
Fixes #299


WasabiFan:
Is it O.K. If you install those deps after installing RVM, or should they be before? I can't remember when if first complains.

dlech:
You just need to install the deps before `bundle install`. They are needed for compiling some of the ruby gems.

WasabiFan:
Ok, this LGTM then

PR #236, opened 2016-09-07
Add upgrade instructions

WasabiFan:
This is an extension of the work by @rikblok to incorporate more detailed information that @dlech suggested in #218. Thanks @rikblok for getting this started!

I tried to make it a balance between "baby steps"-style walkthrough and tl;dr summary. If you want to read the whole thing, you can, otherwise you can just find the heading you want and copy-paste the appropriate command(s).

Preview at <http://wasabifan.github.io/ev3dev.github.io/docs/tutorials/upgrading-ev3dev-from-command-line/>.

_Note: I didn't bother opening up the Brickman desktop test and instead opted to just get a screenshot from one of my running EV3s. It's a bit different from the rest of our screenshots (the IP address and status icons) but other than that it should be functionally the same, including size and color._


docs/tutorials/upgrading-ev3dev-from-command-line.md
```diff
@@ -0,0 +1,74 @@
...
+
+When you flash an SD card with an ev3dev image, you are "installing" the version of ev3dev that was available at the time that the image was made. However, sometimes you will want to upgrade the version of ev3dev that you are using so that you can take advantage of new features that have been added or bug fixes that have been made. One option to get a new version is to re-flash your SD card with a newer image, but this also deletes any files you have saved on it. Alternatively, you can upgrade the ev3dev system (the device drivers and kernel) without having to delete any of your work. In this tutorial, we will show you how to upgrade ev3dev from the command line.
+
+**Make sure that you have already configured a network connection and connected to ev3dev via SSH before beginning.**
+
+
+Before starting the upgrade process, you will need to have ev3dev download a list of the most recent versions of available software. To do this, run the following command:
+
+
+    sudo apt-get update
```
dlech: It would be better if we just listed this with the `upgrade` commands. If it is separate, people will miss it and wonder why upgrade is failing.

docs/tutorials/upgrading-ev3dev-from-command-line.md
```diff
@@ -0,0 +1,74 @@
...
+{% include icon.html type="warning" %}
+The `upgrade` step may take a couple hours to complete, depending on your network
+connection and hardware.
+{: .alert .alert-warning}
+
+Although you don't necessarily need to reboot if no new kernel was installed,
+it's difficult to see whether one was installed based on the command output.
+To reboot, run `sudo reboot` or do it through Brickman.
+
+# Confirming that the upgrade worked
```
dlech: This only confirms that a kernel upgrade have worked. It should be listed immediately after the kernel upgrade section.

docs/tutorials/upgrading-ev3dev-from-command-line.md
```diff
@@ -0,0 +1,74 @@
...
+there was no new kernel available, and you are done (unless you forgot to run
+`sudo apt-get update` as shown above). If the command does not output that line,
+**you will need to reboot the EV3 before the new kernel can be used. Do so through
+Brickman on the EV3 display or by running `sudo reboot`.**
+
+# Upgrading all packages (slower)
+
+Run these commands in the following order:
+
+    sudo apt-get upgrade
```
dlech: need to have `sudo apt-get update` here too.

docs/tutorials/upgrading-ev3dev-from-command-line.md
```diff
@@ -0,0 +1,74 @@
...
+    sudo apt-get dist-upgrade
+
+Wait for each command to complete before typing the next.  Some steps require 
+confirmation to proceed. Press 'y' when asked.  
+
+{% include icon.html type="warning" %}
+The `upgrade` step may take a couple hours to complete, depending on your network
+connection and hardware.
```
dlech: and most importantly, how many packages actually need to be upgraded.

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: One of the feedback items that you had on the original implementation of this PR was that you wanted an explanation of what this command does. I can copy the command into both of the other sections, but where would I then put the explanation?

docs/tutorials/upgrading-ev3dev-from-command-line.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Upgrading Ev3dev from the Command Line
+subject: System
+---
+
+{% include icon.html type="warning" %}
+By upgrading ev3dev you are also upgrading the drivers, so there may be changes that break user software.
```
dlech: You are not necessarily upgrading drivers and drivers are not the only thing that can break other software.

Furthermore, my advice is the Linux Mint philosophy: if everything is working and you are happy, then **don't upgrade**. Only upgrade if A) something is not working for you and you think that upgrading a package will fix the problem or B) you want to help with testing ev3dev and you plan on reporting bugs when you find something that is broken.

docs/tutorials/upgrading-ev3dev-from-command-line.md
```diff
@@ -0,0 +1,74 @@
...
+title: Upgrading Ev3dev from the Command Line
+subject: System
+---
+
+{% include icon.html type="warning" %}
+By upgrading ev3dev you are also upgrading the drivers, so there may be changes that break user software. 
+You should check the ev3dev [news feed](http://www.ev3dev.org/news/){: .alert-link } for breaking changes before upgrading.
+{: .alert .alert-warning}
+
+When you flash an SD card with an ev3dev image, you are "installing" the version of ev3dev that was available at the time that the image was made. However, sometimes you will want to upgrade the version of ev3dev that you are using so that you can take advantage of new features that have been added or bug fixes that have been made. One option to get a new version is to re-flash your SD card with a newer image, but this also deletes any files you have saved on it. Alternatively, you can upgrade the ev3dev system (the device drivers and kernel) without having to delete any of your work. In this tutorial, we will show you how to upgrade ev3dev from the command line.
```
dlech: > sometimes you will want to upgrade the version of ev3dev

We aren't upgrading "versions" we are upgrading "packages"

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: Both of the sections have the goal of upgrading the kernel, it's just a matter of how it is done: target just the necessary packages vs. get upgrades for everything else as well. I think this separate section is important to have so that a) you have obvious instructions for checking regardless of the method (both may have upgraded the kernel) and b) we can link to it directly when telling people to check their kernel version after upgrading.

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: Good point; will-do.

docs/tutorials/upgrading-ev3dev-from-command-line.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+    sudo apt-get update
+
+
+This might take up to ten minutes, depending on hardware and network connection speed.
+
+------
+
+You have multiple options in how you upgrade ev3dev. You can either upgrade just the ev3dev core (which is the quickest option) or upgrade all software installed on the EV3 (which may take an hour or longer).
```
dlech: What is the "ev3dev core"? the kernel? What about upgrading other packages one at a time?

docs/tutorials/upgrading-ev3dev-from-command-line.md
```diff
@@ -0,0 +1,74 @@
...
+Wait for each command to complete before typing the next.  Some steps require 
+confirmation to proceed. Press 'y' when asked.  
+
+{% include icon.html type="warning" %}
+The `upgrade` step may take a couple hours to complete, depending on your network
+connection and hardware.
+{: .alert .alert-warning}
+
+Although you don't necessarily need to reboot if no new kernel was installed,
+it's difficult to see whether one was installed based on the command output.
```
dlech: If there is a `linux-image-*` package listed, then the kernel was upgraded. It's not that hard to see. Also `flash-kernel` will run which is also easy to see.

Well, easy for me anyway. :wink:

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: I'm not sure what you expect me to change based on that information. If people get to this page, they have either found it by themselves or been told to come here. In the former case, they probably know what they are doing anyways (and that case will be rare); in the latter, the context from which they came dictates what they are here for, and it's almost certainly to upgrade ev3dev -- that's what the title says, after all.

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: To a beginner, upgrading "packages" is meaningless. The point of this paragraph is to explain what this notion of "packages" is.

docs/tutorials/upgrading-ev3dev-from-command-line.md
dlech: I think it would make sense to explain all three command (update, upgrade, dist-upgrade) at the end of the "Upgrading all packages" section.

update: fetches the list of packages from the package repository servers
upgrade: upgrades all packages that do not require installing new packages or removing existing packages
dist-upgrade: upgrades packages that require installing new dependencies or removing conflicting packages

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: > What is the "ev3dev core"? the kernel?

The kernel and any packages it depends on.

> What about upgrading other packages one at a time?

That's out of the scope of this tutorial as it stands, and honestly I don't think it should be modified to include that. The individual docs for those packages or forum questions found online will tell the user to upgrade them at various points, but a user won't come to our site asking how to upgrade bash, for example.

docs/tutorials/upgrading-ev3dev-from-command-line.md
dlech: The way I see it, one is upgrading the kernel, the other is upgrading *everything* - which may or may not upgrade the kernel.

docs/tutorials/upgrading-ev3dev-from-command-line.md
dlech: Saying "upgrading ev3dev" is ambiguous. What is "ev3dev" exactly? If you want to make a statement about upgrading drivers, then saying "upgrading the kernel package" would make more sense because upgrading the kernel is specifically what upgrades the drivers.

dlech:
@WasabiFan, I think our main point of contention here is that I am approaching this from the point of view that this page should be about upgrading software packages in general, but you are approaching it from the point of view that this page is about upgrading the kernel.

We need to pick one or the other and make the page a bit more focused on what specifically we are trying to accomplish.

docs/tutorials/upgrading-ev3dev-from-command-line.md
dlech: > The point of this paragraph is to explain what this notion of "packages" is.

How do you do that without ever mentioning the work "package"?

docs/tutorials/upgrading-ev3dev-from-command-line.md
dlech: What about brickman or flash-kernel or python3-ev3dev or any of the other 30+ ev3dev-specific packages?

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: The problem is that the "update" description, as well as the command, is applicable to _both_ sections. I _could_ only include the description in one of the two sections, but then the other would be lacking it.

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: > The way I see it, one is upgrading the kernel, the other is upgrading everything - which may or may not upgrade the kernel.

Agreed. By extension, both could upgrade the kernel, and so you should be able to check if the kernel was upgraded after either of them.

dlech:
And to be honest, most people just want to know "what commands do I have to type in to make it work" they don't want to know the why and the how of it all. This page really could be just 2 paragraphs or so.

Perhaps we need a separate section on the website for the detailed explanations for those that want to learn more about linux. Then we can just leave the tutorials as simple step-by-step instructions without all of the explanations getting in the way.

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: > saying "upgrading the kernel package" would make more sense because upgrading the kernel is specifically what upgrades the drivers.

More technically correct? Yes. More sense to us? Yes. More sense to the average "beginner" user? No. I'm shooting low, and then if someone knows more they can apply it here as well. To most people, they would view "ev3dev" as the whole installation, and associate "ev3dev version" with the generic software version, similar to a version of iOS or WIndows (or even Linux, although they wouldn't have experience with that). It's only people who actually understand what these packages are and how they interact that would make the distinction of "ev3dev version" and "ev3dev kernel version"..

In the end, if someone knows what a "kernel package" is, they probably won't need this page to help them out.

dlech: > To most people, they would view "ev3dev" as the whole installation, and associate "ev3dev version" with the generic software version, similar to a version of iOS or WIndows

I get what you are saying here. But I will argue that this would be analogous to upgrading ev3dev-jessie to ev3dev-stretch. Packages are just "apps" in the iOS or Windows sense. Upgrading the kernel package is like installing Windows updates, not like upgrading Windows itself.

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: Understanding the concept is different from knowing the name; it didn't seem important so I omitted it until it becomes useful. I can add that name if you want, but even then it'd take some creative wording to call it the "ev3dev kernel package" and still have it make sense to a beginner. I might be able to make that work though.

WasabiFan:
>  I think our main point of contention here is that I am approaching this from the point of view that this page should be about upgrading software packages in general, but you are approaching it from the point of view that this page is about upgrading the kernel.

Yeah, that seems right.

> We need to pick one or the other and make the page a bit more focused on what specifically we are trying to accomplish.

Agreed. My goal was to make a page so that we can say something like "You need to upgrade your kernel. To do so, see [here]." I'm afraid of straying too far into the technicalities to make that not straghtforward; in my mind, I am trying to minimize the boilerplate required every time we realize that someone needs to upgrade.

> And to be honest, most people just want to know "what commands do I have to type in to make it work" they don't want to know the why and the how of it all. This page really could be just 2 paragraphs or so.

I think that's what #218 was, more or less.

> Perhaps we need a separate section on the website for the detailed explanations for those that want to learn more about linux. Then we can just leave the tutorials as simple step-by-step instructions without all of the explanations getting in the way.

I like that idea, although I would want to make sure that there is a clear distinction between those sections and that we aren't duplicating significant content between them.

WasabiFan:
And actually, after thinking a bit more, I'd be pretty happy with a set of pages called a "Technical series" (or something similar) which would explain things like packages, sysfs access, etc. for people who want to learn about the technical background.

docs/tutorials/upgrading-ev3dev-from-command-line.md
WasabiFan: I suppose the Win32 API hasn't had any breaking changes since Windows 3.1, so it's more of a difficult comparison that I originally imagined :laughing: We really just need to pick a way to tell users to imagine it and stick with it everywhere.

dlech:
OK. How about this? We'll start a new discussion for the "technical series" (to be continued...)

As for this PR, what if we make it "upgrading ev3dev" instead of "upgrading ev3dev from the command line". The idea is there are 3 ways to "upgrade ev3dev"...

1. flash a new image
2. upgrade everything (remember that the ev3dev-config package can to this too)
3. just upgrade the kernel package

So basically, take what you have here, cut down the explanations to a bare minimum and add a section at the beginning on re-flashing and the pros (probably faster) and cons (wipes out any personal files).

WasabiFan:
Sounds good. What do you think should be done about the image releases? We generally tell people to install the latest nightly; does that mean that this guide should also recommend that (at least temporarily) or do you expect to release a new official image soon?

dlech:
I think that if someone is looking to upgrade the kernel, then 90% of the time, the nightly is what they need (unless we stop using oss.jfrog/artifactory and start posting everything on github again).

PR #237, opened 2016-09-11
Add upgrade instructions (v3)

WasabiFan:
This is a pared-down version of the page that appeared in #236.

I tried to keep it simple, although ev3dev/ev3dev#741 would dramatically simplify option 1. I tried to use bullets instead of paragraphs/sentences to make it easier to skim (because despite our best efforts, people will rightly do it).

As usual, preview [here](http://wasabifan.github.io/ev3dev.github.io/docs/tutorials/upgrading-ev3dev/).

It might make sense to pull out the "checking the version" section into its own page at some point. The page could also mention `uname` and `ev3dev-sysinfo`.


dlech:
:shipit:

WasabiFan:
Sounds good; thanks for the review!

PR #254, opened 2016-09-26
Lidar data photo

bmegli:
Some photos taken when resoldering one of my connectors. Combined, annotated and described step-by-step process in caption.

Also added captions to all other images.

Say NO to your CRC failures...


WasabiFan:
I have to say, I'm impressed by the fact that you managed to produce seven merge commits :laughing:

bmegli:
Wow, I have impressed even myself. In 10 days... but this means less than once a day still ;-)

PR #255, opened 2016-09-26
Testing the lidar with mapping

bmegli:
Detailed instructions how to plot LIDAR readings in real-time with ev3dev-mapping.

No more hardware required (apart from EV3 and XV11) ugh, and possibly USB hub.

Restructured section on Testing The LIDAR


WasabiFan:
6 merge commits 😜

bmegli:
Uhm, yes - this was branched of the same local master that had 7, just before the 7th ;-)

This makes me wonder - is it only me who keeps local repositories in sync with remotes (so often)?

ddemidov:
If you `git fetch && git rebase origin/master` instead of `git pull`, you will have a nice linear history instead of the merge commits.

bmegli:
Thanks @ddemidov.

Yes, I am going to do that, at least for that (knowing when rebasing can cause trouble).

Thank you!

WasabiFan:
As long as the branch that you are on doesn't have any commits that our master branch does not, merging will "fast-forward" without creating a merge commit. So, I usually don't have this problem because I branch off of something that is in the same state as (or behind) master and can pull whenever I need to. It's rare that there are changes I need from master after I have already committed things.

PR #265, opened 2016-10-06
Minor fixes of minor fixes

bmegli:
Let's see if we have minor fixes of minor fixes of minor fixes ;-)


bmegli:
But know that I converted all 3 setups that were verbs to set up

WasabiFan:
Thanks!

PR #270, opened 2016-10-08
Take advantage of new content in getting started

### LINKED ISSUE #245: Emphasize next steps in Getting Started
Make the links to the "Libraries" page from "Getting Started" more obvious and natural as the next step.

This could be taken further by having the "introductory pages" display a "Next step" button at the bottom of the page. This would help with the flow of that series.

From ev3dev/ev3dev#734.

### LINKED ISSUE #250: Remove "calling it a day" section
I think this breaks the flow of the Getting Started guide. This content is valuable, so we may want to break it out into its own page, but the last step in the guide should lead into something about writing code without any barriers in between.

From ev3dev/ev3dev#734.

### LINKED ISSUE #266: Update Getting Started to link to new landing pages (e.g. networking)

WasabiFan:
Fixes #266
Fixes #250
Fixes #245


docs/getting-started.md
```diff
@@ -216,75 +216,51 @@ Troubleshooting tips if your EV3 won't boot
...
-* __Bluetooth__
-    * Note: Bluetooth may not work on the first boot. Please reboot if you see "???"
-      after you power on Bluetooth.
-    * [Connecting to the Internet](../tutorials/connecting-to-the-internet-via-bluetooth) tutorial
+There are lots of choices here. Visit our [networking page](/docs/networking)
+and choose the listed connection option that best suits you and your available
+devices.
 
 {% include icon.html type="info" %}
 For Raspberry Pi with no display, you must use the wired Ethernet port
```
bmegli: Let's assume that display here may meen PiStorms built-in or BrickPi external so it's ok

docs/getting-started.md
```diff
@@ -216,75 +216,51 @@ Troubleshooting tips if your EV3 won't boot
...
 
 <div class="dark-bg colored-section">
 <div class="container" markdown="1">
 
-<div class="row">
-<div class="col-md-8" markdown="1">
-## Step 7: Calling it a day
+## Step 7: Choose a programming language and write some code
 
-When you are ready to wish your EV3 goodnight, turn if off by pressing the
```
bmegli: Don't we want to keep this information somewhere? Or maybe this is just to obvious.

dlech: I don't think it is very obvious.

docs/getting-started.md
```diff
@@ -216,75 +216,51 @@ Troubleshooting tips if your EV3 won't boot
...
-    * [Setting Up Wi-Fi Using the Command Line](../tutorials/setting-up-wifi-using-the-command-line)
-      (requires another connection type first to get to the command line, but it's way
-      easier to enter your passphrase this way)
-* __Bluetooth__
-    * Note: Bluetooth may not work on the first boot. Please reboot if you see "???"
-      after you power on Bluetooth.
-    * [Connecting to the Internet](../tutorials/connecting-to-the-internet-via-bluetooth) tutorial
+There are lots of choices here. Visit our [networking page](/docs/networking)
+and choose the listed connection option that best suits you and your available
+devices.
```
dlech: I think there should be a big button (or something else really noticeable) here for the to make it obvious that clicking on the link is a required step.

Also, I'm no sure linking to the networking page here is an improvement. Now users will have to follow two links to get to the information they need and go back two times to get back to the Getting Started page.

dlech: Perhaps we could list the top 3 options here with direct links and then link to the networking landing page for more options.

docs/getting-started.md
```diff
@@ -216,75 +216,51 @@ Troubleshooting tips if your EV3 won't boot
...
-[here](/docs/programming-languages).**
-
-If the language you want isn't listed, you still can use it, but you'll have to
-do more of the heavy lifting yourself. You can look at the [driver index page](/docs/drivers/)
-for information on the interfaces you need to use to control devices. Once you get the
-hang of it, you can even write your own interface library and have it listed here!
+## Step 6: Connect to the EV3 via SSH
+Now that you have a network connection, you should
+[connect to your EV3 with SSH](/docs/tutorials/connecting-to-ev3dev-with-ssh)
+if you haven't done so already.
```
dlech: I think that we should be going under the assumption here that this is a first time user reading the page. So, we can leave out the "if you haven't done so already".

Also, this section could be made a bit more beginner friendly with a brief explanation that
1. This lets us know that the network connection is working
2. Provides secure remote access to the EV3

We could also give an example command here that does something interesting, such as using `espeak` to say something (will need to think of something else for Raspberry Pi since it does not have a built-in speaker)

docs/getting-started.md
```diff
@@ -216,75 +216,51 @@ Troubleshooting tips if your EV3 won't boot
...
+The brick can run the vast majority of popular programming languages, so your
+favorite language is probably supported.
 
-<div class="col-md-4">
-{% include screenshot.html source="/images/brickman/shutdown-menu.png" caption="The 'Shutdown...' dialog in Brickman" %}
-</div>
-</div>
+<a class="btn btn-md btn-primary" href="/docs/programming-languages"><span class="glyphicon glyphicon-console"></span> See the available programming languages</a>
+
+### Writing your own interface library
```
dlech: I don't think this belongs on the Getting Started page. Very few people have the gumption to actually write their own library - and if they do, they will have no problem finding this information elsewhere. For example, this would be more appropriate as an info box on the programming languages page.

docs/getting-started.md
WasabiFan: Yes, we do want it somewhere else, however I have no idea where that other place would be.

WasabiFan:
Updated.

docs/getting-started.md
ddemidov: May be merge this section with 'Boot ev3dev'? Say, another alert window saying `And by the way, here is how you turn your ev3dev off:`

WasabiFan: I like that idea. It would probably work well as an alert at the end of that section, but depending on how that looks visually I might try something else. I'll test it out and report back.

docs/getting-started.md
vini2710: I think it would be easier if there was just one example that works on both the RPI and the ev3.

WasabiFan: @vini2710 Do you have any ideas for something interesting that would work on both? I liked this because its effects are "obvious" and it feels cool to see the brick speak. Doing something with the display would not be cross platform (nor simple), so that's out as well. Motors would require the hardware and a fair number of statements, so I think that is out too. What does that leave us with?

docs/getting-started.md
```diff
@@ -216,75 +216,58 @@ Troubleshooting tips if your EV3 won't boot
...
-    * Setting up Wi-Fi Via the Brickman User Interface (hopefully you can figure
-      this out, but it would be nice if someone made a tutorial)
-    * [Setting Up Wi-Fi Using the Command Line](../tutorials/setting-up-wifi-using-the-command-line)
-      (requires another connection type first to get to the command line, but it's way
-      easier to enter your passphrase this way)
-* __Bluetooth__
-    * Note: Bluetooth may not work on the first boot. Please reboot if you see "???"
-      after you power on Bluetooth.
-    * [Connecting to the Internet](../tutorials/connecting-to-the-internet-via-bluetooth) tutorial
+There are lots of choices here. Choose the connection option that best suits you
```
dlech: It might be good to say "click on the button below" here since the "lots of choices here" are not actually *here* on this page, they are on another page.

docs/getting-started.md
```diff
@@ -216,75 +216,58 @@ Troubleshooting tips if your EV3 won't boot
...
-If the language you want isn't listed, you still can use it, but you'll have to
-do more of the heavy lifting yourself. You can look at the [driver index page](/docs/drivers/)
-for information on the interfaces you need to use to control devices. Once you get the
-hang of it, you can even write your own interface library and have it listed here!
+<a class="btn btn-md btn-primary" href="/docs/tutorials/connecting-to-ev3dev-with-ssh"><span class="glyphicon glyphicon-transfer"></span> Learn how to connect with SSH</a>
+
+To test this out, let's try running a command. Type (or copy-and-paste) the
+following into the terminal, and press enter. It will use the speakers and
+text-to-speech software to speak out loud. **If you are not on an EV3, this
+will not work. It requires the EV3's built-in speaker.**
```
WasabiFan: This guide currently doesn't support the EVB, so mentioning it for the first time here would seem odd. Based on what we have seen so far with people accidentally downloading EVB images, I would like to avoid modifying any of the language on this page for that platform until we do it all at once with appropriate explanation.

In general, I think that whatever demo we put here should operate without any external hardware, so I avoided explaining that caveat on the RPi side with the hope that we will think of something else for the RPi instead. Should I just modify this here instead and leave the RPi one for later?

dlech: > This guide currently doesn't support the EVB

Sure it does, even if it is not explicitly mentioned.


> In general, I think that whatever demo we put here should operate without any external hardware

Perhaps we should install the `fortune` package by default?

WasabiFan: > Sure it does, even if it is not explicitly mentioned.

That's the point. It isn't mentioned, so we can't expect a user of one of the other platforms to know what it is. Users have demonstrated that not knowing what the EVB is causes issues, so we need to work around that. As it stands, if you are operating under the assumption that EVB users will already be technically-able enough to navigate the guide without specific instructions, they should be able to handle themselves here too.

> Perhaps we should install the fortune package by default?

I like that idea. We won't be able to use it now, though, because this page still refers to the 10-month-old image.

docs/getting-started.md
vini2710: Maybe leds? Not sure if they are controlled in the same way on the EV3, the PiStorms and the BrickPi...

WasabiFan: @dlech suggested below that it might be nice to demo with the [`fortune` program](https://en.wikipedia.org/wiki/Fortune_(Unix)), which prints out a random quote from a set of predefined ones. What do you think about that option? It isn't as cool as doing something visible/audible which is specific to ev3dev, but at least it would work on all platforms and does something interesting.

It's worth noting that that package isn't currently installed in ev3dev images, so we would need to wait until the next release (which is theoretically soon!) before changing the instructions to use it.

docs/getting-started.md
vini2710: >Perhaps we should install the fortune package by default?

Sounds like a good idea!

docs/getting-started.md
vini2710: Sounds good!

dlech:
Can you fix the merge conflicts? Is there any more work to be done on this PR?

WasabiFan:
Yes, there are a few more things I need to do. My hope is that we'll get an image release out which includes the `fortune` package before I finish this so that we don't have to write a description of something temporary in that area.

WasabiFan:
Updated. I believe this is ready for final review and merge.

WasabiFan:
This is live at <http://wasabifan.github.io/ev3dev.github.io/docs/getting-started/>.

bmegli:
Missing dot in line 140

{% include icon.html type="warning" %} This is for Raspberry Pi only! Users of other platforms can skip this step. {: .alert alert-warning }

Should be:

{: .alert .alert-warning }
(dot added before alert-warning)

WasabiFan:
@dlech When you get some time, can you review this and #289?

docs/getting-started.md
```diff
@@ -216,16 +233,17 @@ Troubleshooting tips if your EV3 won't boot
 
 ## Step 5: Set up a network connection
 
-There are lots of choices here. Choose the connection option that best suits you
-and your available devices.
+Click on the button below to visit our networking page. There are lots of
+choices here; choose the connection option that best suits you and your
```
WasabiFan: Yes, it is; does it seem weird?

bmegli: For me - yes, only because we don't use such contructions anywhere else (not because it is incorrect). But I had a feeling "something is not right here?"

If I had to I would rather use hyphen - instead of ; but it is totally subjective and I am non native in English so don't listen ;-)

bmegli: > In general I notice the same patterns from people of the same (or similar) native language.

This calls for machine learning project - predict the native language of a person from it's written English... 😉

WasabiFan:
One thing that I realized hasn't been considered yet is the download routes. The guide currently includes its own download buttons which are separate from the "Downloads" page. Is this what we want?

dlech:
Let's keep the download buttons on the getting started page, but change the paragraph after that to a short sentence with a link to the downloads page. Something like `For more download options, see our [downloads page](/downloads).`

WasabiFan:
Updated.

download.md
```diff
@@ -9,7 +9,7 @@ no-wrapper: true
 <div class="row margin-bottom-xl">
     <div class="col-sm-12">
         <p class="lead">
-            <span class="glyphicon glyphicon-download"/>
+            <span class="glyphicon glyphicon-download"></span>
```
dlech: why change the span elements here and below?

docs/getting-started.md
```diff
@@ -193,77 +192,82 @@ data will be lost. Keep in mind, that it may take a much longer time from 8V to
...
+<div class="col-md-8" markdown="1">
+### Shutting down
+
+You'll often want to turn off your EV3 while it isn't in use. When you are ready
+to wish your EV3 good night, turn if off by pressing the *back* button from the
+main menu in brickman or pressing _and holding_ the *back* button from any screen
+in brickman. This will open a dialog where you can select *Power Off* to turn
+off the EV3. _If you're using the command line, running `sudo poweroff` in a
+terminal works too._
+</div>
```
dlech: `*word*` and `_word_` have the same effect in markdown - *italic*. If you intended **bold**, you need `**word**`, otherwise, let's be consistent and just use `*` everywhere.

dlech: Having a code snippet in italics looks strange too.

docs/getting-started.md
```diff
@@ -193,77 +207,82 @@ data will be lost. Keep in mind, that it may take a much longer time from 8V to
...
+To test this out, let's try running a command. Type (or copy-and-paste) the
+following into the terminal, and press enter. It will use a nifty program called
+`fortune` to print out a random quote from a database. Feel free to try it
+multiple times!
 
-If the language you want isn't listed, you still can use it, but you'll have to
-do more of the heavy lifting yourself. You can look at the [driver index page](/docs/drivers/)
-for information on the interfaces you need to use to control devices. Once you get the
-hang of it, you can even write your own interface library and have it listed here!
+{% highlight bash %}
```
dlech: I don't think bash is a valid highlight name. Try `sh` instead.

docs/getting-started.md
```diff
@@ -193,77 +207,82 @@ data will be lost. Keep in mind, that it may take a much longer time from 8V to
...
+run programs, change settings, and install new programs. Click the button below
+to learn how to connect.
+
+<a class="btn btn-md btn-primary" href="/docs/tutorials/connecting-to-ev3dev-with-ssh" target="_blank"><span class="glyphicon glyphicon-transfer"></span> Learn how to connect with SSH</a>
 
-The brick can run almost all programming languages that any other Linux distro can, so your
-favorite language is probably supported. Language bindings have already been
-written for many languages. **You can learn more about the available libraries
-[here](/docs/programming-languages).**
+To test this out, let's try running a command. Type (or copy-and-paste) the
```
dlech: should probably say "After you have connected using SSH, in the SSH terminal window..." or something to that effect.

docs/getting-started.md
dlech: Nevermind, I was thinking of `shell` from [here](https://github.com/ev3dev/ev3dev.github.io/commit/22c45e1381bfd705b97c5c87e972ce2b3fa625c0#diff-24b8369988afaa96ad38f073e1da2962L50)

download.md
WasabiFan: Because self-closing spans are illegal HTML, and this caused extreme issues in browsers running strict mode.

docs/getting-started.md
WasabiFan: I have no idea what I intended, but I suspect that it's a matter of copy-pasting from multiple authors. What do we use for button text elsewhere?

docs/getting-started.md
```diff
@@ -193,77 +207,82 @@ data will be lost. Keep in mind, that it may take a much longer time from 8V to
...
 {% include icon.html type="info" %}
 For Raspberry Pi with no display, you must use the wired Ethernet port
 to connect for the first time. You can set up additional connections using the
 `connmanctl` command.
 {: class="alert alert-info"}
 
-Now that you have a network connection, you should
-[connect to your EV3 with SSH](/docs/tutorials/connecting-to-ev3dev-with-ssh)
-if you haven't done so already.
+<a class="btn btn-md btn-primary" href="/docs/networking" target="_blank"><span class="glyphicon glyphicon-globe"></span> Check out the available networking options</a>
```
WasabiFan: I imagine a user clicking on things in order, reading the content between each button; if they are doing it that way, they wouldn't see the alert until after they had visited the page and presumably configured a connection. Maybe this alert should just be moved to the networking page; what do you think?

docs/getting-started.md
WasabiFan: IIRC none of them did much, but it was at least detected (I learned from experience that it throws a key-not-found error if it's an invalid language name).

download.md
dlech: Makes sense when you think about what a span is. I guess htmlproofer is just not a very good validator. I've found other things that are obviously incorrect html that it has missed.

docs/getting-started.md
dlech: In most places, I have been using `*thing*`.

docs/getting-started.md
dlech: moving the alert sounds good to me

WasabiFan:
Updated. I can't rebuild the preview site at the moment though.

docs/getting-started.md
```diff
@@ -257,29 +251,37 @@ to learn how to connect.
...
+To test this out, let's try running a command (make sure that you have completed
+the earlier part of this step first).
 
-{% highlight bash %}
-fortune
+Type the command `fortune` into the SSH shell opened above and press
+<kbd>Enter</kbd>. It will use a nifty program called `fortune` to print out a
+random quote from a database. Feel free to try it multiple times!
+
+{% highlight none %}
```
dlech: Apparently, `none` is no good. If you really want no highlighting, then just indent 4 spaces and get rid of the highlight liquid tags.

WasabiFan: Ummmm... it works locally 😕

WasabiFan: Replaced with indentation

dlech: Have you run `bundle update` lately?

WasabiFan: I just (forcefully) reinstalled Ruby and re-cloned the repo last week, so everything should be as new as our gemfiles allow.

dlech:
Anything else you want to change before we merge?

WasabiFan:
Nope, I think this is good-to-go.

PR #277, opened 2016-10-12
Download

dlech:
I need to get images for everything, so I am not quite done yet, but have a look....

http://dlech.github.io/ev3dev.github.io/download/


WasabiFan:
Nice! A few early suggestions (I'm judging with temp images, so I can't say how it will look in the end with those added):
- The light background from the jumbotron seems weird on a page with its own title. I think you could drop the jumbotron, and instead maybe use our background light/dark classes from the getting started guide to make the 2nd section light-colored.
- The EV3 item looks lonely because it's so basic 😆 For the purpose of making it a bit bigger (and adding background info to a lesser extent), adding a sentence or two saying how the EV3 is our original and primary platform might be nice. It's also worth considering if we want to do anything to draw attention specifically to this option, as it is the most common.
- The double images are confusing to my eye. The page doesn't feel visually balanced because there are bookending columns of images with left-aligned text between them, not to mention the fact that the first row doesn't have a second image. The second image also makes it look a bit crowded.

  I like the left-aligned text with the single image a lot, but I'm unsure of how the second can be incorporated as well. I'm tempted to do an accordion-like fold of the images so that it unfolds when you hover over it (and the whole thing expands to the right) similar to GitHub's notifications:

  fold

  I would much prefer to find a static solution though. Do you have any ideas? Another option might be to put the text and buttons on the left side and the images stacked leftward floating on the right. That might be a good choice.
- The top half of the page looks odd in mobile view. I'd suggest investigating how it looks to right-alignin the images in that view.
- Do we want to link to release notes as well?
- Remember to add a link in the nav and update the getting started page.

WasabiFan:
Despite my best efforts, the Travis build is still unhappy with other links... let's hope that sorts itself out.

dlech:
> The light background from the jumbotron seems weird on a page with its own title. I think you could drop the jumbotron, and instead maybe use our background light/dark classes from the getting started guide to make the 2nd section light-colored.

Thanks. I knew we had something like this around somewhere.

> The EV3 item looks lonely because it's so basic :laughing: For the purpose of making it a bit bigger (and adding background info to a lesser extent), adding a sentence or two saying how the EV3 is our original and primary platform might be nice. It's also worth considering if we want to do anything to draw attention specifically to this option, as it is the most common.

I don't like using text as "filler" but yes, it does seem like it needs something.

> The double images are confusing to my eye. The page doesn't feel visually balanced because there are bookending columns of images with left-aligned text between them, not to mention the fact that the first row doesn't have a second image. The second image also makes it look a bit crowded.

I added more images just to see what it looks like. I don't particularly like it either at the moment.

> Do we want to link to release notes as well?

People probably won't read them anyway, but yes this would be a good thing to do. We will need to automate this somehow though.

vini2710:
>Another option might be to put the text and buttons on the left side and the images stacked leftward floating on the right. That might be a good choice.

Maybe you could test that?

WasabiFan:
@dlech Could you move the commits which add new helpers and other unrelated things to their own PR? I'd prefer to keep this PR localized so that the diff is easier to understand, and that has the added advantage that we can separately review the other changes.

> I added more images just to see what it looks like. I don't particularly like it either at the moment.

Hmmm... Maybe this should be reorganized to use the platform core as the heading instead of the EV3 compat boards. I would expect that anyone using BrickPi, for example, knows that they are running it on a RasPi, so you could put "Raspberry Pi" as the heading for that section and mention the specific boards below. That would mean that you could have one image per row in a way that makes sense.

I just tested out the design that I speculated about before, where the images are all on the right side. I just used the dev tools to move the first column to be after the one with the text in it in each row. This is how it looks:

image

With some tuning (e.g. image sizes and row alignment) I think it could look pretty good.

> People probably won't read them anyway, but yes this would be a good thing to do. We will need to automate this somehow though.

The simplest approach would probably be to guess the release note file name based on the image name, but I have no idea how reliable that would be. The best option would probably be to add a data file which correlates image names with release note files, going forward from here. We'd have to update that file each time we release, but at least it's something that can be done in advance before a release is published.

dlech:
> Could you move the commits which add new helpers and other unrelated things to their own PR?

Some of them. This PR depends on most of them though. You can click on the individual commits to comment on each one if that helps.

I'm going to hold of on release notes for now. Since we changed to using docker to create images, we currently are not automatically generating files anyway.

dlech:
Updated: http://dlech.github.io/ev3dev.github.io/download/

WasabiFan:
The row coloring looks good!

I think the platform list would look a bit better if you removed the margin on the left. Also, it would be great if the pictures were cropped so that they were all the same size.

The fact that there are multiple major headings in the RasPi and BB sections makes it look odd. It would be nice if there were a single heading for each with a sub-heading list below it.

download.md
```diff
@@ -0,0 +1,153 @@
+---
+title: Downloads
+subtitle: Select the SD card image file for your hardware below
```
WasabiFan: We generally haven't referred to the page itself in the subtitle, so I think this would be better without the "below"

dlech:
Updated again: http://dlech.github.io/ev3dev.github.io/download/

WasabiFan:
Looks great! The only significant issue I see (which I only just thought of) is that we should add a note about the getting started guide to prevent people from habitually clicking the "download" link and then not knowing where to go. A note at the top or below the downloads section saying that you should make sure to follow the guide should be sufficient.

Two notes:
- The name for the RasPi device is "Raspberry Pi Zero", not "Raspberry Pi 0"
- Can you update the "upgrading ev3dev" page to use the same table as the one you have here? It's a lot nicer than the one I made originally.

dlech:
I added a link to the Getting started page.

> The name for the RasPi device is "Raspberry Pi Zero", not "Raspberry Pi 0"

I'll remember this if I am ever writing it out in prose. But I opted to keep the short hand 0 here for symmetry on the buttons.

Issue #155, opened 2016-05-12
link verification does not always work for https sites on travis


dlech:
Searching google for [html-proofer travis ssl](http://lmgtfy.com/?q=html-proofer+travis+ssl) shows we aren't the only ones that are having issues.

The latest is https://www.hiawatha-webserver.org/. We are also already ignoring https://warmcat.com because of ssl issues.

These work locally, but not on travis for whatever reason.

WasabiFan:
My (barely) educated guess is that the sites that fail our build are running an older version of TLS/SSL. Or maybe it's a newer version that is the problem. I'm not really sure.

I politely asked Travis to run a simple cURL on one of the failing URLs and got this:

```
$ curl --url https://www.hiawatha-webserver.org/ --verbose
* About to connect() to www.hiawatha-webserver.org port 443 (#0)
*   Trying 141.138.201.249... connected
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
* SSLv3, TLS handshake, Client hello (1):
* SSLv3, TLS alert, Server hello (2):
* error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
* Closing connection #0
curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version


The command "curl --url https://www.hiawatha-webserver.org/ --verbose" exited with 35.
```

dlech:
It seems to be working now. I've restarted some of the checks for pull requests and they are green.

WasabiFan:
Yeah, I just tested it and it's working again for me too. I'd say this can be closed (at least for now).

dlech:
https sites keep breaking :cry:

PR #278, opened 2016-10-12
fix failing links

dlech:
This fixes the kernel.org links. OpenRoberta is still failing, but they just upgraded their server, so it may or may not resolve itself.

I'm wondering if we should just stop checking links with travis altogherter since it can't seem to handle https very well.

### DIFF
--- a/cibuild.sh
+++ b/cibuild.sh
@@ -36,6 +36,7 @@ IGNORE_HREFS=$(ruby -e 'puts %w{
     robosnap.net
     01.org
     alldatasheet.com
+    kernel\.org
 }.map{|h| "/#{h}/"}.join(",")')
 
 # Explanation of ignored sites:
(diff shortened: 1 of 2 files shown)


WasabiFan:
> I'm wondering if we should just stop checking links with travis altogherter since it can't seem to handle https very well.

The CI server has caught so many broken links that I would feel pretty insecure without it. When I originally enabled link checking, I had to fix 50-100 real broken links; the hacker docks, for example, had major issues. LEGO has also been repeatedly breaking links and rarely adding proper redirects. Despite the fact that it can be a pain to special-case the HTTPS endpoints, I still think it provides significant benefit.

It would probably be a good step to reach out to Travis support and/or open an issue on the proofer library to see if there's a real solution to this.

There are, however, workarounds we can use to make this significantly better. I currently have it configured to run nightly builds of `master` and have been watching those as they happen to make sure that things are running smoothly. I also wrote [a Ruby script](https://github.com/WasabiFan/ev3dev.github.io/commit/a122bc8445ead696256ea19d5f19e3d95eb1fadc) that I was trying out a few months back which would only fail the build for errors in modified files, but I didn't get a chance to test it out thoroughly enough to be confident in it.

dlech:
Reopening #155 for now...

PR #279, opened 2016-10-12
Rename "Get Help" to "Support"

dlech:
This is more conventional and in the menu bar it is better to have
all items as nouns, not a mix of nouns and verbs.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -44,7 +44,7 @@
                         </ul>
                     </li>
                     <li>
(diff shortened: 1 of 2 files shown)


WasabiFan:
By the way, I'm intentionally just approving these instead of merging them outright so that I don't accidentally mess up your other PRs in case there is a specific procedure you wanted to use to minimize weirdness.

dlech:
For future reference, you can go ahead an merge (unless I specifically say "do not merge yet" in the comments). I know how to rebase. :wink:

PR #285, opened 2016-10-15
Use separate list for IGNORE_HREFS on travis

dlech:
Travis has issues with https, so we have to ignore quite a few sites.

This lets us actually check the sites ignored by travis locally every
once in a while while keeping travis green.


dlech:
See #155

PR #287, opened 2016-10-16
Add platform comparison page

dlech:


WasabiFan:
Can we get some HTML formatting?

dlech:
There is some at the bottom. :grin:

dlech:
And there is a preview: http://dlech.github.io/ev3dev.github.io/docs/platform-comparison/

WasabiFan:
😵 that's a fancy table. That might be the most complicated table I've ever seen.

> There is some at the bottom. 😁

Is there something preventing an auto-formatter from formatting it?

dlech:
I tried using a fancy html editor, but I guess it was not that fancy. should be better now.

WasabiFan:
Ah, that's much more readable!

What's happening with the multi-column cells here (and in a few other places)? Is it just because there were two next to each other which said the same thing? I would prefer it if they were separate cells.

image

Could you try making the border on the first column more visible to differentiate the headings from the values?

I think this page might benefit from a sentence of intro/explanation.

dlech:
> What's happening with the multi-column cells here (and in a few other places)? Is it just because there were two next to each other which said the same thing?

Yes

> I would prefer it if they were separate cells.

I *knew* you were going to say that.

> Could you try making the border on the first column more visible to differentiate the headings from the values?

Any css tips for this?

dlech:
I also need to add a link from the main docs page.

WasabiFan:
> I knew you were going to say that.

I'm not actually OCD, but I've found myself to be a bit of a perfectionist :laughing:

> Any css tips for this?

Something like this, probably:

```css
.table-index-column {
    border-right-color: white;
    border-right-width: 3px;
}
```

Then you'd apply that class to each cell in the column. There may be a Bootstrap class for this already, but I doubt it; you'll probably just have to fiddle with the color and weight.

WasabiFan:
You may want to wrap this monolithic table in a `<div class="table-responsive">` to make it scrollable on mobile.

dlech:
:new:

docs/platform-comparison.html
```diff
@@ -0,0 +1,286 @@
+---
+title: Platform Comparison
+---
+
+<p>
+    This table provides a detailed comparison of all of the ev3dev supported
```
WasabiFan: Hyphenate "ev3dev-supported"

docs/platform-comparison.html
```diff
@@ -0,0 +1,286 @@
...
+    devices and their features.
+</p>
+
+<div class="table-responsive">
+    <table class="table table-striped table-bordered table-left-header">
```
WasabiFan: Nice thinking with the header class

docs/platform-comparison.html
```diff
@@ -0,0 +1,286 @@
...
+            <td>LEGO</td>
+            <td colspan="2">Dexter Industries</td>
+            <td>mindsensors.com</td>
+            <td>FatcatLab</td>
+            <td>Quest Institute</td>
+        </tr>
+        <tr>
+            <th colspan="2">Website</th>
+            <td><a href="http://mindstorms.lego.com" target="_blank">mindstorms.lego.com</a></td>
+            <td colspan="2"><a href="http://www.dexterindustries.com/BrickPi" target="_blank">dexterindustries.com</a></td>
```
WasabiFan: What's this "target" stuff?

dlech: open in a new page

dlech:
:new: :new:  http://dlech.github.io/ev3dev.github.io/docs/platform-comparison/

dlech:
I tried `.lead` and it looked out of place with no regular text afterwards, so leaving as-is for now.

PR #288, opened 2016-10-18
Image release

dlech:
http://dlech.github.io/ev3dev.github.io/news/2016/10/17/ev3dev-jessie-2016-10-17-release/


WasabiFan:
Looks good! The only thing I'd say is that if you are going to mention including the output from `ev3dev-sysinfo`, we might want to mention the `-m` option for that purpose. Also, we need to update our issue template.

WasabiFan:
For this type of thing, I think we should standardize on squashing on merge to combine the "oops" commits and add a detailed commit message.

PR #290, opened 2016-10-20
Add missing "div.row"s to project and news pages

WasabiFan:
GitHub's diff is pretty ugly, but all this does is wrap these sets of columns into `div.row`s to make things aligned properly. At some point these were removed/changed but I have no idea when (nor do I really care).


dlech:
LGTM

PR #298, opened 2016-11-18
Update ev3-bluetooth.md

dlech:
It seems that TI has removed the CC2560 product page. At least the wiki page is still up.


WasabiFan:
I submitted a feedback item to TI about this; I'm hoping that they can at least be a bit more cognizant of page removals/relocations.

dlech:
This part has been out of production for quite a while...

WasabiFan:
I don't really care if they're no longer producing the part; it's evident that the part still exists in the wild, and until that ceases to be the case it is irresponsible of them to remove information on it from their website. At a minimum, they should have a message explaining why the page is gone, and preferably would have a link to the next-best option.

WasabiFan:
From TI support:

> We apologies [sic] for the inconvenience that this has affected you. This was caused by an error in our cache which made some pages available. There are some system errors which causes [sic] pages to be unavailable from time to time. Please rest assured that we are monitoring on the website and will take immediate action when an issue occurs.

The pages in question here are now available again, so if they're valuable you should be able to change the links back to what they were before this PR. I can't say their technical explanation is particularly enlightening, but I'll take it.

dlech:
Cool. Thanks!

PR #304, opened 2016-12-03
Structure includes folder and prioritize essential resources

WasabiFan:
It turns out that our page load time is actually pretty bad, with the large number of individual resource files that we have. This should improve the impression of page load time by only blocking rendering for critical CSS; everything else is loaded after the rest of the page is parsed. These changes are most intelligible as two separate commits because I also added some folders to the "includes" folder.


dlech:
> It turns out that our page load time is actually pretty bad,

define "bad".

The extra level of organization for includes looks good to me. All of the pages with a `.markdown` extension are automatically generated, so we need to also change [sensor-defs-to-markdown.py](https://github.com/ev3dev/ev3dev-kpkg/blob/ev3dev-jessie/sensor-defs-to-markdown.py#L211) and friends so that these changes don't get reverted during the next kernel release.

WasabiFan:
> define "bad".

Around three seconds on average according to Google Analytics (excluding unexplained outliers), with roughly 20% of users waiting between three and seven seconds and 8% waiting more than that *. These changes don't actually decrease the amount of time spent loading things, but it makes it so that the page will _appear_ to have loaded much faster because the browser is loading the JavaScript and other extras in the background after the above-the-fold content is rendered. If we really want to decrease time to all assets loaded, we'll need to figure out how to get Jekyll to minify and combine our files, which I am not prepared to do at this point.

> All of the pages with a .markdown extension are automatically generated, so we need to also change sensor-defs-to-markdown.py and friends so that these changes don't get reverted during the next kernel release.

Whoops, I completely forgot about that since I made my original reorg commit. I'll open a PR over there when I get the chance.

---

_* I can't actually exclude the odd outliers in those later statistics, so there might be some skewing there depending on what caused them. If it was just a single user whose browser left the page unloaded for a few minutes, that wouldn't significantly impact the stats._

dlech:
That's pretty bad.

PR #307, opened 2016-12-10
Configure kramdown to use GFM

WasabiFan:
This makes it so that code fences work properly in local previews. I believe that the GH Pages build already default to this setting, because fences worked on the real site before, but not in previews.


WasabiFan:
Code fences allow us to specify a code highlighting language, so we now will be able to normalize on code fences instead of using the ugly `highlight` blocks.

PR #312, opened 2016-12-20
Format news archive with dates and less ugly titles

### LINKED ISSUE #283: News archive page should show dates for articles.
The [news archive page](http://www.ev3dev.org/news/archive/) should show dates in addition to the title. I always use the date of blog posts to help estimate how useful they might be.

Also, the titles don't need to be quite so big on this page.

WasabiFan:
Fixes #283

### DIFF
--- a/news/archive/index.html
+++ b/news/archive/index.html
@@ -6,10 +6,12 @@ title: News Archive
 
 {% for post in site.posts %}
     {% if post.categories contains "news" %}
-    <div class="news-item">
-        <div class="news-item-header">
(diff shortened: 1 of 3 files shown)


dlech:
Do you have a screenshot of what this looks like? `<h1>` sounds rather large for not having any content in between the titles.

WasabiFan:
Whoops, sorry; I guess I forgot to post my screenshot! The size of the heading was being overridden by a CSS rule, but in reality it shouldn't be an `h1` anyway, so I switched it out for an `h2`. This is the same formatting that appears on the main news page, just without the content of the panel.

image

PR #314, opened 2016-12-22
ev3dev-jessie-2016-12-21 announcemnt

dlech:


WasabiFan:
@JorgePe Is `http://ofalcao.pt/` your site? If so, it seems to be having issues with its _base de dados_ 🤣

dlech:
I left you the `e` because the announcement is so short. Just kidding. I just made that up.

JorgePe:
@WasabiFan yeap... everytime someone finds my site, mysql goes down. I need to tune it better or give up wordpress. Thanks.

WasabiFan:
Haha sorry about that... I suppose us linking to a site is pretty much the equivalent of a DOS attack for small set-ups. If it would be helpful, we can whitelist your site so that our builds just assume that those links are valid.

JorgePe:
The whitelist would be great now... this morning I updated my VPS to a more recente ubuntu version without noticing a major mysql upgrade... now my wordpress blog is down, I'm waiting for service desk to restore a previous backup.

PR #317, opened 2016-12-22
Whitelist links to ofalcao.pt

WasabiFan:
This site is @JorgePe's blog, and is (or was) having some intermittent server issues. This will lower the strain on those servers.

### DIFF
--- a/cibuild.sh
+++ b/cibuild.sh
@@ -43,6 +43,7 @@ if [ "$TRAVIS" == "true" ]; then
         lab\.open-roberta\.org
         questforspace\.com
         bountysource\.com
+        ofalcao.pt
     }.map{|h| "/#{h}/"}.join(",")')
 else
     # credit: code snippet borrowed from jekyllrb.com website source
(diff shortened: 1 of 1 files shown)


JorgePe:
Thanks. Site is up again, new major versions of PHP and MySQL and minor versions of Apache, hope it can handle a bit more at least while I read some performance tuning for LAMP servers.
Am I the only one with external links? I feel flattered :)
(being the only one with a lousy private server is not so flattering)

dlech:
I don't think this will make much difference. On a busy day, we *might* check the links 10 times, but is certainly shouldn't be a significant load on any web server no matter how puny it is.

WasabiFan:
For a server that otherwise gets a few clicks a week, us hitting it that many times is some serious load :laughing: In general, it shouldn't make much of a difference though. It's really up to @JorgePe -- do you think you'll get benefit out of us not hitting your site ~4 times a day on average?

JorgePe:
Not much.
I have ~50 visitors each day and can withstand much more (in theory, at least).
My recent problems are related to mysql respawning to fast, usually when one of my youtube videos gets some attention (this week was my MQTT video, probably because you're revising my tutorial but also because there is another nice EV3 MQTT video on youtybe and that probably brings me more visitors).
So I think that the problem is not average traffic but some spikes.
I've changed a MySQL parameter that improved endurance but it still happens ~ once per month.
If things don't get better, I'll have to change my blog from wordpress to something more stable.

WasabiFan:
OK, I'm going to close this then, because it sounds like your site is back up-and-running. Feel free to ask if you'd like us to reopen!

PR #322, opened 2016-12-27
"Humanize" breadcrumbs and improve misc formatting

WasabiFan:
This is mainly targeting formatting of titles, however I threw in a bunch of other minor things from my run-through of almost all pages:
- Wrapped lines
- Updated "share" page now that mindsensors.com has removed their poll
  - Re-formatted that page and added a "star" button using an unofficial library which brings you to our repo (note that it doesn't actually star upon click)
- Switched text on "support" page to reference ev3dev-lang-python, as ev3dev-lang is now useless for users and Python is the most common library in use.
- Updated "support" instructions now that we have an issue template
  - _Side-note: I think it would me nice to figure out how to shrink the "IRC" section, as it makes the Gitter section hard to spot and is the less-optimal option._
- Updated breadcrumbs to make them more human-readable

Demo is live here: http://wasabifan.github.io/ev3dev.github.io/docs/drivers/dc-motor-class/

### DIFF
--- a/_includes/page-core/breadcrumbs.html
+++ b/_includes/page-core/breadcrumbs.html
@@ -7,24 +7,24 @@
     {% for dir in parent_dirs %}
         {% assign link=link | append: dir | append: '/' %}
 
-        {% assign humanized_name = dir %}
+        {% assign pretty_name = dir %}
         {% if forloop.last %}
             {% assign itemElem = 'span' %}
             {% assign liClass = 'active' %}
-            {% assign humanized_name = page.title %}
+            {% assign pretty_name = page.title %}
         {% else %}
             {% assign itemElem = 'a' %}
             {% assign hrefAttrib = 'href="' | append: link | append: '"' %}
 
             {% assign current_dir = current_dir[dir] %}
             {% if current_dir.name %}
-                {% assign humanized_name = current_dir.name %}
+                {% assign pretty_name = current_dir.name %}
             {% endif %}
         {% endif %}
 
         <li property="itemListElement" typeof="ListItem" class="{{liClass}}">
             <{{itemElem}} {{ hrefAttrib }} property="item" typeof="WebPage">
-                <span property="name">{{ humanized_name }}</span>
+                <span property="name">{{ pretty_name }}</span>
             </{{itemElem}}>
             <meta property="position" content="{{position}}">
         </li>
--- a/_includes/page-core/breadcrumbs.html
+++ b/_includes/page-core/breadcrumbs.html
@@ -2,19 +2,29 @@
 {% assign link='/' %}
 <ol class="breadcrumb" vocab="http://schema.org" typeof="BreadcrumbList">
     {% assign position = 1 %}
+    {% assign current_dir = site.data.breadcrumbs %}
+
     {% for dir in parent_dirs %}
         {% assign link=link | append: dir | append: '/' %}
+
+        {% assign humanized_name = dir %}
         {% if forloop.last %}
             {% assign itemElem = 'span' %}
             {% assign liClass = 'active' %}
+            {% assign humanized_name = page.title %}
         {% else %}
             {% assign itemElem = 'a' %}
             {% assign hrefAttrib = 'href="' | append: link | append: '"' %}
+
+            {% assign current_dir = current_dir[dir] %}
+            {% if current_dir.name %}
+                {% assign humanized_name = current_dir.name %}
+            {% endif %}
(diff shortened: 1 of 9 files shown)


_includes/page-core/breadcrumbs.html
```diff
@@ -2,19 +2,29 @@
 {% assign link='/' %}
 <ol class="breadcrumb" vocab="http://schema.org" typeof="BreadcrumbList">
     {% assign position = 1 %}
+    {% assign current_dir = site.data.breadcrumbs %}
+
     {% for dir in parent_dirs %}
         {% assign link=link | append: dir | append: '/' %}
+
+        {% assign humanized_name = dir %}
```
dlech: I think I would rather call this `pretty_name` instead of `humanized_name`. It is quite common to see the term "pretty name" in other toolkits, so I already have a good idea of what it means.

share.md
```diff
@@ -1,29 +1,30 @@
...
-<h3><span class="glyphicon glyphicon-star heading-icon"></span>Give us a star on GitHub</h3>
-"Stars" are a great way to show support for a GitHub project. Visit [ours][] and star us!
-</div>
-    <div class="col-md-6" markdown="1">
-<h3><span class="glyphicon glyphicon-check heading-icon"></span>Vote for ev3dev</h3>
-Vote for ev3dev in the [mindsensor.com][] poll (it's about half way down
-the page).
-</div>
-</div>
+Give us a star on GitHub
```
dlech: If this is supposed to be a heading, I would rather use `## Give us ...` instead of the underline (`---`)

share.md
```diff
@@ -1,29 +1,30 @@
...
+    data-count-href="/ev3dev/ev3dev/stargazers"
+    data-count-api="/repos/ev3dev/ev3dev#stargazers_count"
+    data-count-aria-label="# stargazers on GitHub">Star</a>
 
-<hr/>
+"Stars" are a great way to show support for a GitHub project. Visit [ours][ev3dev-repo] and star us!
 
 Show us your ev3dev projects
-------------------------
+---
```
dlech: Same here. `## Show us ...` would be better.

support.md
```diff
@@ -4,9 +4,6 @@ redirect_from: /issues/index.html
 excerpt: "Have a problem or question? We are here to help - but you have to help us help you. We keep track of problems, suggestions and questions about ev3dev using GitHub Issues. This lets us keep everything in one place."
 ---
 
-* Table of Contents
```
dlech: Why delete the table of contents? This page is rather long.

support.md
```diff
@@ -69,15 +67,15 @@ In the description of your issue, be sure to also include:
     `systemctl status brickman.service -l`.
   * If a screenshot of the EV3 would be helpful, run `fbgrab <some-name>.png`
     on the EV3 and include the picture in your comments.
-<p />
```
dlech: Is there not a matching `<p>` to delete? Or is this another case of "WTH<sup>*</sup> HTMLProofer, I thought you were suppose to *validate* things".

<sub>* "what the heck" is my preferred exclamation.</sup>

share.md
WasabiFan: I think I probably just normalized to whatever was below it; if the prefix version is preferred (which is what I generally use) I'll switch both out for that.

support.md
WasabiFan: It was rendering incorrectly in local builds for this particular page and I didn't see an obvious benefit. Looking at it more closely, I think it would be beneficial to keep it, so I'll see if I can get it working properly.

dlech:
One thing I just realized is that for many of the driver pages, the name of the driver is not listed anywhere other than in the breadcrumbs. This should be on the page somewhere. I'll have to check if we need to update the scripts that automatically generate the pages or if there is something in a template that we can update.

support.md
WasabiFan: I'm afraid it is the latter :laughing: I am not particularly well-learned in this area, so I can't say how this is viewed in general use. I know that only _void elements_, which can never include content (such as `br`), can be self-closing in HTML5; however that might not have been the case for XHTML or HTML4.

dlech: OK. Make that WTH @dlech. I read `</p>` instead of `<p />`, which does not compute in my brain.

WasabiFan:
PR updated. I can't figure out that ToC issue -- it only happens on that page in Edge. It seems like none of the scripts, including the ToC styling hook, are being run 😕 Either way it seems functional when live.

Issue #331, opened 2017-01-20
Broken Links


dlech:
This is going to be a perpetual issue since we seem to almost always have broken links. The idea is this is a way to communicate that someone has done something to notify webmasters that their site is broken. For example, a day or two ago, I emailed fatcatlab to let them know their site was down and they were able to fix it.

dlech:
The latest breakage is docker. This has already been reported in their github issues 10+ times, so no need to report it again.

https://github.com/docker/docker.github.io/issues/1195

WasabiFan:
What do you think about adding the ability to specify temporarily-broken sites in an environment variable through the Travis web interface? That would allow us to get builds up-and-running in the short term when things like this happen.

dlech:
Interesting idea. I suppose it would be nice for new contributors who are trying to get a green light. Of course, it would make it more effort to find out when links are fixed though.

dlech:
Broken links of the day: 

- ./docs/kernel-hackers-notebook/ev3-lcd/index.html

  *  External link https://plus.google.com/+JamesKasper/posts/EtTj24hBao3 failed: 404 No error

  *  External link https://plus.google.com/108244279740379322507/posts/KfWiuzGRRKx failed: 404 No error

htmlproofer 3.4.0 | Error:  HTML-Proofer found 2 failures!


These pages are up, so no action taken yet.

dlech:
google plus seems to have fixed itself.

New breakage:

*  External link https://www.hiawatha-webserver.org/ failed: response code 0 means something's wrong.
    It's possible libcurl couldn't connect to the server or perhaps the request timed out.
    Sometimes, making too many requests at once also breaks things.
    Either way, the return message (if any) from the server is: SSL connect error

possible candidate for excluding because of ssl errors

WasabiFan:
This one's fun:
```
- ./docs/tutorials/setting-up-an-nfs-file-share/index.html
  *  External link http://www.barryodonovan.com/index.php/2012/12/12/apple-os-x-as-an-nfs-server-with-linux-clients failed: 301 Peer certificate cannot be authenticated with given CA certificates
htmlproofer 3.4.0 | Error:  HTML-Proofer found 1 failure!
```

WasabiFan:
[...]

PR #343, opened 2017-02-12
Link uses https now

dlech:
#331


WasabiFan:
🎉 `response code 0 means something's wrong.`

PR #344, opened 2017-02-15
Add tutorial for reusing ssh connections

Dave-Baum:
Are you able to squash my commits as you merge the pull request?  Since I edited this directly on github, every save turned into a commit, so there are far more commits than are useful.


dlech:
> Are you able to squash my commits as you merge the pull request?

Yes, we can do that, so no problem there.

docs/tutorials/reuse-ssh-connections.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Reuse SSH Connections
```
dlech: All (or most) of our other tutorials use the "ing" form of the verb, so "Reusing SSH Connections" would match better with other tutorials. The file name should be changed to match as well.

docs/tutorials/reuse-ssh-connections.md
```diff
@@ -0,0 +1,59 @@
...
+## Congifuration (OSX, Linux)
+
+Edit or create a new file at `~/.ssh/config`.  Add the following contents:
+
+    Host <name>
+      HostName <address>
+      User <user>
+      ControlMaster auto
+      ControlPath ~/.ssh/master-%r@%h:%p
+
```
dlech: Is the extra indentation meaningful for the lines after the first? I don't see anything in the `ssh_config` man page about this.

docs/tutorials/reuse-ssh-connections.md
```diff
@@ -0,0 +1,59 @@
...
+Where `<name>` is the name you want to use to refer to your ev3, `<address>` is its ip address, and `<user>` is
+the user name (the default is `robot`).  For example, if your ev3 has address `192.168.1.10`, you want to
+call it `ev3` and connect as user `robot`, the file would look like this:
+
+    Host ev3
+      HostName 192.168.1.10
+      User robot
+      ControlMaster auto
+      ControlPath ~/.ssh/master-%r@%h:%p
+
```
dlech: ev3dev ships with avahi, so mDNS resolution works out of the box with macOS and Linux. I think this would be sufficient for most people:

    Host ev3dev.local.
    User robot
    ControlMaster auto
    ControlPath ~/.ssh/master-%r@%h:%p

Then you just connect with `ssh robot@ev3dev.local.`

dlech: Although having a shorter alias for `robot@ev3dev.local.` can be nice too. So maybe this could be the works-for-eveyone copy-and-paste example:

    Host ev3
    HostName ev3dev.local.
    User robot
    ControlMaster auto
    ControlPath ~/.ssh/master-%r@%h:%p

And just mention as a note that if this gives you trouble, try using the IP address instead of `ev3dev.local.`

docs/tutorials/reuse-ssh-connections.md
```diff
@@ -0,0 +1,59 @@
...
+## Intro
+
+Establishing an ssh connection requires a a lot of communication between your host computer
+and the EV3 and can take several seconds.  Although this is fine a long-lived connection,
+it can be quite tedious if you want to use ssh to invoke a single command or scp to transfer
+files.  By instructing ssh to reuse connections, each connection after the first one will
+reuse the session that was already established, drastically speeding up ssh and scp commands.
+
+## Congifuration (OSX, Linux)
```
dlech: "macOS" is the new "OSX". We use both on the site, so it doesn't matter too much, but I have been trying to standardize on "macOS" since that is now the official name.

dlech: Do you also use Windows? I think you can do something similar with PuTTY, so it would be nice to have a section for Windows users as well.

WasabiFan: There are some letters mixed up here: "Con *gif* uration"

docs/tutorials/reuse-ssh-connections.md
```diff
@@ -0,0 +1,59 @@
...
+
+Edit or create a new file at `~/.ssh/config`.  Add the following contents:
+
+    Host <name>
+      HostName <address>
+      User <user>
+      ControlMaster auto
+      ControlPath ~/.ssh/master-%r@%h:%p
+
+Where `<name>` is the name you want to use to refer to your ev3, `<address>` is its ip address, and `<user>` is
```
WasabiFan: As @dlech mentioned below, you can use the hostname instead of the IP address.

docs/tutorials/reuse-ssh-connections.md
```diff
@@ -0,0 +1,59 @@
...
+      User robot
+      ControlMaster auto
+      ControlPath ~/.ssh/master-%r@%h:%p
+
+Note that if you already have a DNS entry for your ev3, you can use that name in the `Host` line and you do
+not need a `HostName` line.
+
+## Testing
+
+Start by creating a new ssh session to your EV3.  If it was named `ev3`, you would do the following:
```
WasabiFan: Replace double spaces with single

docs/tutorials/reuse-ssh-connections.md
```diff
@@ -0,0 +1,59 @@
...
+
+    ssh ev3
+
+After a few seconds you will be prompted for a password and then will be greeted with the shell
+prompt from the ev3.
```
WasabiFan: "ev3" -> "EV3" in sentences

docs/tutorials/reuse-ssh-connections.md
dlech: Since there is the possibility of adding a Windows section to this tutorial, I would rather keep the OS out of the title/filename. And actually, this will already work on Windows as-is if you are using, for example, OpenSSH in Cygwin.

WasabiFan: Currently, this tutorial doesn't call out the supported platforms at all; it needs to be clarified somehow. An alert doesn't make much sense, and including it in a single subheading is also illogical. With that, it seems to me like including the OS in the title is the best option. The OS shouldn't be included in the file name though.

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: Sorry, but I haven't touched a Windows machine in years.  If someone else can fill in the details that would be great.  But for something like this I have no idea how much of the generic *nix stuff carries over to PuTTY, etc.

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: Indentation is ignored.  I'm used to seeing configs that use the indentation to improve readability, but it isn't necessary so I'll remove it.

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: Cool - I didn't know it broadcasted an mDNS name.

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: What about adding a Windows section as TBD?  The best end state here is to actually have details for Windows, and to me it doesn't make sense for a page to change its title as we fill in more details.

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: Done

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: This section was changed to take advantage of mDNS.

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: Done

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: Done

docs/tutorials/reuse-ssh-connections.md
Dave-Baum: What's the verdict on this?  If you want a different title and/or filename, let me know what they should be changed to.

docs/tutorials/reusing-ssh-connections.md
```diff
@@ -0,0 +1,51 @@
...
+Edit or create a new file at `~/.ssh/config`.  Add the following contents:
+
+    Host ev3
+    HostName ev3dev.local.
+    User robot
+    ControlMaster auto
+    ControlPath ~/.ssh/master-%r@%h:%p
+
+This configuration does several things: it establishes `ev3` as a nickname (so you don't have to type
+`ev3dev.local.` all the time), defaults to `robot` as the user name, and  instructs ssh to automatically
```
WasabiFan: There's an excess dot after `ev3dev.local`

WasabiFan: And below

dlech: It actually works both ways. The trailing dot means that it is an absolute domain name rather than relative. It doesn't really matter for this tutorial, so we can leave it as-is (as of a814064).

dlech:
Thanks!

PR #346, opened 2017-02-18
Update connecting-to-the-internet-via-usb.md

pmclee:
Added link to instructions to follow in case Windows doesn't automatically load the NDIS driver for the EV3+ev3dev device.

While following the steps on this page my windows 7 machine was not able to load the NDIS driver. It took me several minutes to debug the problem and find a solution.

### DIFF
--- a/docs/tutorials/connecting-to-the-internet-via-usb.md
+++ b/docs/tutorials/connecting-to-the-internet-via-usb.md
@@ -156,6 +156,9 @@ If you are using an older version, please upgrade.
 
         {% include /util/screenshot.html source="/images/windows/10/devices-and-printers-rndis.png" %}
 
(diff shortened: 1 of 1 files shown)


WasabiFan:
We had instructions of our own for those steps on this page which were removed in https://github.com/ev3dev/ev3dev.github.io/commit/4b33c0d9185c0c8f96a7b70d6129a78879a7ca45. @dlech will need to comment on the technological basis for these changes.

dlech:
We made a change in USB networking in the last release that could have broken things. I tested that it was working on Windows 7, so if it is not working for some people, I think we should dig a little deeper. A workaround should not be needed. Please open an issue at https://github.com/ev3dev/evdev/issues and state the exact steps to repeat the problem.

PR #347, opened 2017-02-18
Add information for Debian and derivatives

bidaian:
In Debian at least, the package brigde-utils needs to be installed manually. Without it, there is no pan1 network bridge and the bnep connection will fail, without indicating what the problem is.

### DIFF
--- a/docs/tutorials/connecting-to-the-internet-via-bluetooth.md
+++ b/docs/tutorials/connecting-to-the-internet-via-bluetooth.md
@@ -190,7 +190,8 @@ If you are using an older version, please upgrade.
     {: .alert .alert-info}
 
     1.  The default Bluetooth and network settings in Ubuntu do not have support for
-        this, so first you need to install `blueman` on your host computer.
+        this, so first you need to install `blueman` on your host computer. Make sure
+        you have the `bridge-utils` package too, it's required.
 
             sudo apt-get install blueman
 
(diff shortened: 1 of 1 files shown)


dlech:
do you have Network Manager installed in Debian. This PR is probably fine, but I am just trying to understand what is different that you need an extra package installed.

bidaian:
I followed the tutorial and it didn't work. As soon as I installed bridge-utils everything went smoothly. It took me like two days of trying to debug the issue, reading docs, ...

I decided that I could save somebody else's time, so before I created the pull request I uninstalled bridge-utils from the server, rebooted both systems (server and robot), removed the pairing and paired them again, and tried to connect the robot, but I had the same error I had before ("bnep failed" on the client, "Can't add bnep0 to the bridge pan1: No such device" on the server).

I installed bridge-utils again and then everything worked at it should (logged in the server as: "bluetoothd: bridge pan1: interface bnep0 added").

Maybe the person that tested Ubuntu already had the package. What I can confirm is that, without it, in Debian Jessie, it does not work. It makes sense to need it to, because that's how the internet connection is shared, via a bridge.

If you want to change the wording, maybe just suggest to install it in case there are problems, then it's ok. Or just a hint, however you see fit.

You could also ask whoever did the testing in Ubuntu to check if they have the package, and if things work once it is removed... (I don't know of any program depending on it).

bidaian:
I tried again and now it works without bridge-utils. I tried so many things that I just don't know what made it work. Sorry for the noise, you were right to hesitate. Closing.

dlech:
Thanks for double-checking. 😄

PR #348, opened 2017-02-18
Fix broken link so that Travis doesn't fail

bidaian:
The information on bitmap fonts moved in the site. This is the updated link.


WasabiFan:
Thanks!

PR #351, opened 2017-03-10
Add icon denoting external nav docs link

WasabiFan:
image

### DIFF
--- a/_includes/page-core/page-header.html
+++ b/_includes/page-core/page-header.html
@@ -35,7 +35,10 @@
                                 <a href="/docs/networking">Networking</a>
                             </li>
                             <li>
-                                <a href="/docs/driver-overview">Hardware Drivers</a>
+                                <a href="http://docs.ev3dev.org/projects/lego-linux-drivers/en/ev3dev-jessie/">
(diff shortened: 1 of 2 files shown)


_includes/page-core/page-header.html
```diff
@@ -35,7 +35,10 @@
                                 <a href="/docs/networking">Networking</a>
                             </li>
                             <li>
-                                <a href="/docs/driver-overview">Hardware Drivers</a>
+                                <a href="http://docs.ev3dev.org/projects/lego-linux-drivers/en/ev3dev-jessie/">
+                                    Hardware Drivers
+                                    <i class="fa fa-external-link" id="external-link-nav-icon" aria-hidden="true"></i>
```
dlech: Aren't ids supposed to be unique per page? It seems like you want a class rather than an id here.

WasabiFan: I had intended for it to be a unique ID, but upon consideration it seems like it might be useful in other links... we'll see if we decide to relocate other pages.

dlech:
Awesome! Thanks!

I'll let you squash and merge however you like.

PR #354, opened 2017-03-14
Replace defunct MinuteBot link

sethten:
The minutebot domain appears to be down. This has been causing Travis CI to fail on every build, and GitHub to mark commits with "check failed." Replace the link with the most recent archive of that page.


WasabiFan:
I mentioned this here: https://github.com/ev3dev/ev3dev.github.io/issues/331#issuecomment-285843734

Pinging @bmegli

bmegli:
@sethten thanks for fixing that

Apart from site down it seems that minutebot base is discontinued. I am not sure about that yet.

PR #355, opened 2017-03-15
Hide redirected pages from search and use firendly breadcrumbs

### LINKED ISSUE #323: Consolidate breadcrumb logic and update breadcrumbs in site search
Currently, the search breadcrumbs use the old-style text instead of the new "pretty" version.

WasabiFan:
GitHub's super unhappy with this diff 😀 

I took advantage of time with this file to also do some slight minification; there's significantly less whitespace in there now. I'd guess it is 30% smaller just from whitespace, and the removed docs pages bring that to something more like a 80% reduction.

Fixes #323

### DIFF
--- a/search-index.json
+++ b/search-index.json
@@ -19,23 +19,29 @@ layout:
         "href":"{{ page.url }}",
         "loopid":"page-{{ forloop.index }}",
         "category":"{{ page.categories }}",
-        {% capture _ %}
+        {% capture _ %} Capturing the generated content will prevent whitespace from being rendered
+
         {% assign parent_dirs=page.url | replace_first: '/', '' | split: '/' %}
         {% assign current_dir = site.data.breadcrumbs %}
         {% assign breadcrumbs = '' %}
+
         {% if parent_dirs.empty %}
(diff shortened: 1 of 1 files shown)


dlech:
Also, GitHub magic to ignore whitespace changes: https://github.com/ev3dev/ev3dev.github.io/pull/355/files?w=1

WasabiFan:
Yeah... Liquid isn't the most _readable_ language :grin:

WasabiFan:
I added some comments; it doesn't really help, but it's probably as good as it will get.

rhempel:
It's not the most readable language because it's barely a writable language - but if it's good enough to generate language bindings from xml I can live with it.

WasabiFan:
> Thanks. I guess you don't need to put them in {% comment %} blocks because of the {% capture %}?

Yeah, it's throwing away the generated content there so I can make it a bit cleaner. It feels dirty but it works :wink:

> It's not the most readable language because it's barely a writable language - but if it's good enough to generate language bindings from xml I can live with it.

It's impressive what we've been able to do with this language!

PR #364, opened 2017-04-14
Kernel release cycle 20

dlech:


WasabiFan:
Is the fix for https://github.com/ev3dev/ev3dev/issues/847 included as well?

dlech:
Yes, I fixed that at the last minute and forgot to add it to the list.

PR #366, opened 2017-04-22
Add ball catcher project

dsharlet:
I'm about 2 years too slow with this... better late than never? :)


dlech:
Thanks Dillon. This is one of my all-time favorite EV3 robots.

PR #372, opened 2017-05-16
Change main-menu.png

markyi370:
I think this picture looks nicer. I can shrink it if you want.


dlech:
FYI, we can't use the LEGO logo on our website. (See https://www.lego.com/en-us/legal/legal-notice/fair-play).

It would be helpful if you could post a screenshot or do a preview so we can see the difference in context.

PR #377, opened 2017-07-17
Update setting-up-an-nfs-file-share.md

ivankeller:
Add `sudo` in front `systemctl` commands on EV3 to avoid Permission denied error.


dlech:
Thanks!

PR #380, opened 2017-07-26
Win10 connecting-to-the-internet-via-bluetooth.md

enitsche:
insert information for  Windows 10

PR #381, opened 2017-08-01
Images for Win10 connecting-to-the-internet-via-bluethooth

enitsche:
deleted the old pictures and loaded them up again. (With new names.)


dlech:
It would be better if these were added to #380 instead of making a separate pull request.

Issue #387, opened 2017-11-18
Section "Previewing your changes online" is unclear


JesApp:
I think the section  "Previewing your changes online" of `README.md` has two problems:

1. It refers to `./publish <gh-user>` which should be `./publish.rb <gh-user>`.
1. The fact that it's a separate section from "Previewing locally (which contains all the ruby setup) suggests - at least to me - that ruby is not needed for previewing online. Unless I missed something, that's not true. Maybe reorder to make the ruby setup it's own section.

WasabiFan:
Those both make sense to me. I think the latter problem is a result of the fact that the online preview originally didn't require Ruby, and we changed the methodology after originally writing the guide. Would you be able to open a PR for this? (You can make the edits directly from the GitHub website.) I think adding a sentence referencing the earlier section would be fine for the second issue.

JesApp:
Thanks for your answer. I made a pull request. I don't know which parts of the ruby setup are relevant for the online preview, though. It's probably better to discuss that over in the [PR](https://github.com/ev3dev/ev3dev.github.io/pull/388).

JesApp:
The PR has been merged, closing.

PR #388, opened 2017-11-19
clarify section on online preview

JesApp:
This PR originated [here](https://github.com/ev3dev/ev3dev.github.io/issues/387).

One open question:

Which parts of the ruby setup in section "Previewing locally" are actually needed for the online preview? Just Ruby itself or is Jekyll needed as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Previewing Your Changes Online
 
 We have a special script to use to publish your changes using the `gh-pages`
 branch of your fork of ev3dev.github.io. If you have SSH setup for your GitHub
-account, simply run`./publish <gh-user>` where `<gh-user>` is your actual GitHub
+account, simply run`./publish.rb <gh-user>` where `<gh-user>` is your actual GitHub
 user name.
 This will copy the contents of `_site`, fix it up a bit and push it to your
 fork on GitHub. You can view the results at `http://<gh-user>.github.io/ev3dev.github.io`.
--- a/README.md
+++ b/README.md
@@ -63,13 +63,7 @@ refresh your browser.
 Previewing Your Changes Online
 ---
 
-We have a special ruby script to use to publish your changes using the `gh-pages`
+After you have previewed your changes locally, you can use a special Ruby script to publish your changes using the `gh-pages`
 branch of your fork of ev3dev.github.io. 
 
-It requires you to install Ruby (see previous section), but Jekyll is not needed.
-
-If you have SSH setup for your GitHub
-account, simply run`./publish.rb <gh-user>` where `<gh-user>` is your actual GitHub
-user name.
-This will copy the contents of `_site`, fix it up a bit and push it to your
-fork on GitHub. You can view the results at `http://<gh-user>.github.io/ev3dev.github.io`.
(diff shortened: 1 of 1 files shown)


dlech:
You just need ruby itself for running the publish script.

But I guess that would explain why most people don't do a preview if I ask... because it is too much work to install ruby on Windows. So, another option would be to turn the publish script into a bash script. Windows users should already have git installed if they manged to clone the repository, so they should have git bash available to run the script.

dlech:
On the other hand, much of our documentation is moving elsewhere for ev3dev-stretch, so maybe we should just leave the publish script as it is.

JesApp:
I agree. I'm on linux and I won't install ruby plus whatever else just preview a markdown file. A bash script would be a much better solution if that's doable. 

But yeah, if you're making big changes in the near future, I'd say leave it as it is. I just used emacs' standard markdown->html export and that worked well enough for a preview.

README.md
```diff
@@ -63,9 +63,13 @@ refresh your browser.
...
 
-We have a special script to use to publish your changes using the `gh-pages`
-branch of your fork of ev3dev.github.io. If you have SSH setup for your GitHub
-account, simply run`./publish <gh-user>` where `<gh-user>` is your actual GitHub
+We have a special ruby script to use to publish your changes using the `gh-pages`
+branch of your fork of ev3dev.github.io. 
+
+It requires you to install Ruby (see previous section), but Jekyll is not needed.
+
+If you have SSH setup for your GitHub
```
dlech: You don't actually have to have SSH setup. The default is to use https and ask for a password. I think if you run `./publish.rb <gh-user> --ssh` then it will use SSH instead.

README.md
```diff
@@ -63,9 +63,13 @@ refresh your browser.
...
+account, simply run`./publish.rb <gh-user>` where `<gh-user>` is your actual GitHub
```
dlech: need a space between `run` and ``` ` ```

JesApp:
Just added the info that previewing online requires Ruby as well. Not the nicest way of doing this, but if large changes are coming anyway, separating things into a "Install Ruby" and a "Install Jekyll and preview locally" section seems a bit overkill to me. I'll remove the [WIP].

README.md
```diff
@@ -63,9 +63,13 @@ refresh your browser.
 Previewing Your Changes Online
 ---
 
-We have a special script to use to publish your changes using the `gh-pages`
-branch of your fork of ev3dev.github.io. If you have SSH setup for your GitHub
-account, simply run`./publish <gh-user>` where `<gh-user>` is your actual GitHub
+We have a special ruby script to use to publish your changes using the `gh-pages`
+branch of your fork of ev3dev.github.io. 
+
+It requires you to install Ruby (see previous section), but Jekyll is not needed.
```
dlech: Technically, the `bundler` gem is not needed either.

README.md
dlech: At least, that is how it is supposed to work.

README.md
WasabiFan: Why are the dependencies not needed? The publish script just publishes the most recent local build, so you need to run a build as usual first.

dlech: Ah, you are right. I had forgotten. I was thinking that jekyll ran on GitHub just like when we publish the real website.

So, please disregard my previous comment. So, I suppose that we should just say that "after you preview locally, you can publish it for other to see using the `./publish.rb` script"

JesApp: Thanks for the info, worked it in.

dlech:
Thanks!

PR #389, opened 2017-12-15
Elaborate on Step #7

tcwan:
Expand on the choice of programming languages and development steps.
This needs to be merged in conjunction with changes to ev3dev repository for source images.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -284,7 +284,19 @@ skip this step and return there for further instructions.
 The brick can run the vast majority of popular programming languages, so your
 favorite language is probably supported. Pick your language and write some code!
 
+1. Choose a Language
+
 <a class="btn btn-md btn-primary" href="/docs/programming-languages"><span class="glyphicon glyphicon-console"></span> See the available programming languages</a>
+
+2. Choose a Toolchain
+
+The following flowcharts can help you select the appropriate toolchain for:
+* [C/C++]
+* TBD
+
+3. Build Libraries (TBD)
+
+4. Download to Target & Debug/Run your program (TBD)
 
 {% include /style/end-section.html %}
 
@@ -294,3 +306,4 @@ favorite language is probably supported. Pick your language and write some code!
 [microSDXC]: https://en.wikipedia.org/wiki/Secure_Digital#SDXC
 [releases]:https://github.com/ev3dev/ev3dev/releases
 [nightly]:https://oss.jfrog.org/list/oss-snapshot-local/org/ev3dev/brickstrap/
+[C/C++]:https://github.com/tcwan/ev3dev/images/workflow-c-cpp.svg


WasabiFan:
Many languages don't have a concept of a tool chain for our purposes, mainly because they're interpreted/JITted. Perhaps this information should live in the repository docs for C and C++ libs; what do you think? What's the main use case you are looking to address with this info?

I appreciate the work you're doing and the critical eye to improvements across all our areas!

dlech:
I agree with @WasabiFan that the Getting Started page is not the best place for this since most people don't ever need to think about toolchains. How about we just put this on the wiki for now and once it is streamlined and polished up, I think it will be more clear to us where it will fit.

tcwan:
Ok. I'll start by drafting the info in the Wiki then.

But Is there a way to specify a URL to refer to a diagram in the repository which points to (i) a fixed version (ii) the latest version canonically without the long hashed URLs?

dlech:
* specific version, not using the git hash: use a git tag, e.g. https://github.com/tcwan/ev3dev/raw/ev3dev-jessie-2017-09-14/logo/ev3devLogo128.png
* latest version: use a git branch, e.g. https://github.com/tcwan/ev3dev/raw/tcwan-wiki-swarch-1/images/workflow-c-cpp.svg 

You can also clone the wiki as a git repository and upload images that way. You probably won't have permission to push to the ev3dev organization's wiki via git, but you can push to your copy of the wiki and I can merge it.

tcwan:
> You can also clone the wiki as a git repository and upload images that way. You probably won't have permission to push to the ev3dev organization's wiki via git, but you can push to your copy of the wiki and I can merge it.

I'm not sure how to clone the ev3dev-wiki. 
The wiki repository does not show up in the ev3dev repository as a subproject??

From the wiki page, Github only allows me to clone it locally to my system, which would imply that my clone is not hosted on Github. Consequently I'm not sure if you can merge it unless I create a new project in my Github account for the wiki.

dlech:
> I'm not sure how to clone the ev3dev-wiki.

   git clone https://github.com/ev3dev/ev3dev.wiki.git

https://help.github.com/articles/adding-and-editing-wiki-pages-locally/

Then you can push to your fork of the ev3dev repository.

    cd ev3dev.wiki
    git remote add tcwan https://github.com/tcwan/ev3dev.wiki.git
    git push -f tcwan master

dlech:
Or you can just edit the ev3dev wiki directly and I can copy the images later if that is easier for you. 😃

tcwan:
>Then you can push to your fork of the ev3dev repository.
>`cd ev3dev.wiki`
>`git remote add tcwan https://github.com/tcwan/ev3dev.wiki.git`
>`git push -f tcwan master`

That was what I thought I should do but I wasn't sure if it's what you meant.
Thanks for clarifying.

tcwan:
I found out that I cannot create the wiki repository manually. Instead I had to do it from my fork of the ev3dev repository and then replace its contents with those from the original ev3dev.wiki.

In addition, checkins using https:// don't work for me (it'll always reject my user/pass), I have to use git:// checkins. 

Anyway, all this is sorted out now, I'll work on the actual contents and update here when something is in reasonable shape.

tcwan:
Sorry to bug you guys constantly, but I'm pretty new to the whole Github.io and Markdown formatting process. I tried embedding URLs into the SVG but I think that Github sanitizes the SVG file to remove embedded links, and lead to missing boxes in the flowchart.

This is what I'm getting currently:
<https://github.com/tcwan/ev3dev/wiki/Getting-Started-Developing-on-ev3dev>

If I remove the embedded URLs then the missing boxes render correctly, but that reduces the functionality of the guide tremendously. 

In addition, I'm not sure if this will affect content of any pages hosted on Github.io or only the Wiki, and if there is a good way to whitelist the embedded URLs.

WasabiFan:
Where exactly are these URLs? If you're just rendering an SVG in the wiki, its content shouldn't be modified by GitHub at all... it's just your browser rendering the SVG file. So I would be inclined to say this sounds like a general problem with the SVG rather than something GitHub-specific.

dlech:
https://stackoverflow.com/questions/13808020/include-an-svg-hosted-on-github-in-markdown

tcwan:
Sounds like I'm really hitting the edge-cases here trying to embed URLs in SVGs.
Sigh. If there's no resolution in the short term I'll have to resort to alternative ways of presenting the information (without linking to URLs from the SVG).

tcwan:
Hi, Happy New year to everyone!

I've be filling in the details for the content in my fork of the Wiki:
https://github.com/tcwan/ev3dev/wiki/Getting-Started-Developing-on-ev3dev

Please take a look

PR #393, opened 2018-07-08
pictures for loev3go

obo:
I'm adding a thumbnail which will hopefully emerge in the projects page and two pictures to my post page.


dlech:
thanks!

PR #396, opened 2018-08-23
Change docker-engine to docker-ce in tutorial

laurensvalk:
Older versions of Docker were called docker or docker-engine.


dlech:
Thanks!

PR #405, opened 2018-10-22
Create 2018-02-17-Lego-Clock.md

riddy8051:
This project is written in shell script on a lego EV3 brick running Debian EV3dev Linux. On boot the EV3 brick gets the current time from the internet via WiFi. When the clock script is started the Ev3 brick finds the hands of the clock with a colour sensor and sets the clock hands to the current time. It keeps reading the current time and moves the clock hands accordingly. A medium motor is geared down 3:1 to drive the minute hand then a further 12:1 (36:1 total) to drive the hour hand. Enjoy!


riddy8051:
Hi dlech,
Thanks for putting it on the site.  Sorry about the minor problems with the post, this is my first entry into the world of Github!

PR #406, opened 2018-12-19
Move JavaScript library to "out-of-date" section on languages page

WasabiFan:
NodeJS isn't supported anymore so the library isn't either.

Also updated Python link URL.


jabrena:
Maybe, it is necessary to do the same with a `Java` library with no support for Debian Stretch.

PR #410, opened 2019-05-01
Added Zero and 3 on buttons

artpoz:
Added Rasperry Pi Zero and Rasperry Pi 3 on download buttons in "Getting started" page


dlech:
Thanks!

PR #412, opened 2019-05-05
Update .travis.yml  - added package to avoid SSL error

artpoz:
Added package to avoid SSL error
Source
https://stackoverflow.com/questions/41522256/how-can-i-avoid-travis-ci-ssl-errors


dlech:
It looks like this does not actually fix the error

Issue #414, opened 2019-07-01
Add Prolog API to list of languages


maweki:
We have created a Prolog API for ev3dev that we use in teaching. As the work has now been published, we want to make our API available to the world:

https://github.com/maweki/ev3dev-prolog

As it turns out, programming a robot using Prolog works surprisingly well, as was predicted 30 years ago (and not often translated into practive). If it is preferred, I can make a pull-request with the link.

dlech:
Cool. Yes, please make a pull request.

One tip for your README, be sure to add `sudo apt update` to the install instructions. Many people don't know that you have to do that first or `apt install` will fail.

maweki:
This is now done.

PR #415, opened 2019-07-02
Added Prolog API

maweki:
As per #414, added a (possibly incomplete) implementation written in Prolog.

### DIFF
--- a/docs/programming-languages.md
+++ b/docs/programming-languages.md
@@ -79,6 +79,14 @@ C-based code. It is very lightweight and often the most portable across platform
 
 [Get started with C](https://github.com/in4lio/ev3dev-c){: .btn .btn-default }
 
+## Prolog
+<https://github.com/maweki/ev3dev-prolog>
+
+Prolog is a logic programming language that is often used in the context of AI
+and planning. Prolog programs allow for a high level of abstraction and a mix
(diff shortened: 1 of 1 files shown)


dlech:
Before I merge, I just wanted to make sure you put this in the *Out-of-date, abandoned and unfinished implementations* section on purpose. If this project is active, it should be moved up before *Vala, Genie and other GObject-based languages with ev3devKit*.

maweki:
It is active and I even hope to find students to work on it further but it is incomplete. For example, there is no support for audio output, display, dc-motors, etc..

For the latter we did not have the available hardware to test this and for the former we did not have the (wo)manpower to implement it. I am happy to put it in the first category and I am happy to support the library for the foreseeable future but it is by no means complete. It has just enough features for our AI courses combining Prolog planning algorithms with actual robotics hardware.

You decide where it fits best. I am happy to move it.

dlech:
If it runs on ev3dev-stretch and works with motors and at least a few sensors, I would call it good enough to bump up to the main list.

maweki:
Thank you. It's good that it is on the homepage as I will be presenting this work on at least two (academic) conferences in Europe this year. I hope this finds some interest.

dlech:
Merged, thanks!

PR #417, opened 2019-10-28
Fix syntax highlighting

cnrmck:


dlech:
I'm afraid this would actually break the highlighting. The current website looks OK to me.

https://www.ev3dev.org/docs/tutorials/using-lego-dimensions-toy-pad/

cnrmck:
@dlech Ooh, my bad. I landed on this in GitHub from a StackOverflow question, unaware of the website. When I viewed it in the GitHub markdown it definitely did not work, but now I understand why. 

Sorry to waste your time :/

PR #418, opened 2019-11-19
Add information for configuring a RaspberryPi

DanielHabenicht:
There is no link to the other guide and searching the web ends in min 2 hours of wasted time. So I added a convenient link.

### DIFF
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -21,7 +21,7 @@ connected, you can then set up additional connections using
 
 # Connecting to the Internet
 
-You have a few options for connecting to the internet.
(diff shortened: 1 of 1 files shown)


dlech:
Thanks for the suggestion. There is already a big blue box on this page that mentions the Raspberry Pi and contains the same link. The link was probably easy to miss since it was formatted as `code`.

I've change the page so that the link is more obvious and added a second link in the Wi-Fi section in case people ignore the big blue box.

DanielHabenicht:
ohh my, yeah now I see it too. Maybe it is better to just incorporate it in the text beneath. 

I've just skipped to the first heading, as it starts with `For Raspberry Pi platforms with no display, you must use the wired Ethernet...` (and that was not was I was looking for). You're changes aren't live yet but I think this should catch all cases. Thanks!

PR #422, opened 2020-03-28
Fix BluPants.md

blupants:
Replaced the template text by real content.

Apologies for the back and forth. I did not have the Ruby env to test the final result. I removed the template data from the 2020-03-25-BluPants.md file and replaced it by real content about my project. Everything should look good now.

Thank you for your patience.


dlech:
merged, thanks!

PR #428, opened 2020-07-26
Fixed broken link

mattruzzi:
Changed mindstorms.lego.com to lego.com/mindstorms

### DIFF
--- a/docs/platform-comparison.md
+++ b/docs/platform-comparison.md
@@ -18,7 +18,7 @@ platforms and their features.
             </tr>
             <tr>
                 <th colspan="2">Website</th>
-                <td><a href="http://mindstorms.lego.com" target="_blank">mindstorms.lego.com</a></td>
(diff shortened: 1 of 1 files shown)


dlech:
Thanks. Although EV3 is retiring soon, so maybe we should pick another link that will have something about EV3 for longer in the future?

mattruzzi:
Should we link directly to the ev3 set page?  <https://www.lego.com/product/lego-mindstorms-ev3-31313>
There also is a [different link for the education edition](https://education.lego.com/products/lego-mindstorms-education-ev3-core-set/5003400), so we could also make the model numbers below link to the page for that model number.

dlech:
I think linking to the LEGO Education site sounds like a good idea. It is hard to say how long the link for the 31313 set will be around after the new MINDSTORMS comes out.

dlech:
merged, thanks!

PR #434, opened 2021-04-29
Add new project

tassosblackg:
Add new project back from 2017 my custom made vehicle (undergraduate at ECE,TUC)


dlech:
Thanks! This is nice to have since people ask about the HiTechnic motor controller from time to time.

PR #435, opened 2021-05-14
Update setting-up-an-nfs-file-share.md with helpful command

GingerIndustries:
Added helpful command related to installing nfs on certain inux systems


dlech:
LGTM, thanks!

PR #436, opened 2022-01-20
Update connecting-to-ev3dev-with-ssh.md

mdoryammilwalrus:
Add instructions for Windows built-in SSH client


dlech:
Thanks for the PR. I would rather keep this all on one *Windows* tab. PuTTY still works on Windows 10.

PR #442, opened 2024-04-08
Mention how to update apt

rayluo:
Per [discussion here](https://github.com/ev3dev/ev3dev/issues/1626#issuecomment-2032101387).

I ended up modifying Getting Started page only, as the download page already references the former.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -244,6 +244,15 @@ available hardware.
 
 {% include /style/begin-section.html %}
 
+## Step 5.1: Update `apt`
+
+Currently, the image file downloaded from step 1 was built in 2020,
+some of its built-in packages are dated, and `apt` does not work out of the box.
+
+To fix `apt` in general, `/etc/apt/sources.list` needs to be changed to match
+https://github.com/ev3dev/docker-library/blob/master/ev3dev-stretch/layers/debian/etc/apt/sources.list
+
+
 ## Step 6: Connect to the EV3 via SSH
 
 Now that you have a network connection, you should connect to your EV3 with SSH.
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -244,6 +244,15 @@ available hardware.
 
 {% include /style/begin-section.html %}
 
+## Step 5.1: Update `apt`
+
+Currently, the image file downloaded from step 1 was built in 2020,
+some of its built-in packages are dated, and `apt` does not work out of the box.
+
+To fix `apt` in general, `/etc/apt/sources.list` needs to be changed to match
+https://github.com/ev3dev/docker-library/blob/master/ev3dev-stretch/layers/debian/etc/apt/sources.list
```
dlech, line 253: Could we suggest how to do this for less experienced users? For example `sudo nano /etc/apt/sources.list` may be the simplest way.

docs/getting-started.md
```diff
@@ -244,6 +244,15 @@ available hardware.
 
 {% include /style/begin-section.html %}
 
+## Step 5.1: Update `apt`
```
dlech, line 247: I think this needs to be after step 6 since an SSH connection is required to update the `sources.list` file.

docs/getting-started.md
```diff
@@ -244,6 +244,15 @@ available hardware.
...
+
+Currently, the image file downloaded from step 1 was built in 2020,
+some of its built-in packages are dated, and `apt` does not work out of the box.
```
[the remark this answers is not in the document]
rayluo: In my next commit, I can remove the "some of its built-in packages are dated, and" sentence. Will that work?

dlech:
Looks good, thanks!

PR #444, opened 2024-05-21
Improve two web links

maehw:
See the teeny-tiny changes in my two commits.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -278,7 +278,7 @@ you're good to go!
 Currently, the image file downloaded from step 1 was built in 2020.
 `apt` does not work out of the box.
 
-To fix `apt` in general, `/etc/apt/sources.list` needs to be changed to match [SourcesList]
+To fix `apt` in general, `/etc/apt/sources.list` needs to be changed to match [this file][SourcesList]
 
 Specifically, you can run `sudo nano /etc/apt/sources.list`,
 delete the old content from that file,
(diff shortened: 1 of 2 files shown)


docs/getting-started.md
```diff
@@ -278,8 +278,7 @@ you're good to go!
 Currently, the image file downloaded from step 1 was built in 2020.
 `apt` does not work out of the box.
 
-To fix `apt` in general, `/etc/apt/sources.list` needs to be changed to match
-https://github.com/ev3dev/docker-library/blob/master/ev3dev-stretch/layers/debian/etc/apt/sources.list
+To fix `apt` in general, `/etc/apt/sources.list` needs to be changed to match [SourcesList]
```
dlech, line 281:
```suggestion
To fix `apt` in general, `/etc/apt/sources.list` needs to be changed to match [this file][SourcesList]
```

"SourcesList" is OK as a lookup key but seems a little odd as the actual text to display.

maehw:
Hi @dlech ,

thanks for your very fast review and the suggested change.

Actually I was in a kind of tunnel and not really aware that I was editing a Markdown file... or in my understanding some generator for static HTML would run on this and I was just trying to adapt the syntax, being unaware that this was not just using tokens.

I've so far only been aware of the following kind of URL syntax:

* `[description](url)`
* what I'd call "inline HTML": `<a href="url" target="..." [other options]>description</a>`

I've committed your suggestion using GitHub's web interface.

Thanks!

dlech:
Looks like it worked as expected. Thanks!

PR #445, opened 2024-11-03
Update connecting-to-ev3dev-with-ssh.md

tim-band:
-o PreferredAuthentications=password is necessary if your ssh client is set to try other methods first. We could instead offer this as a response to the error message "Received disconnect from 192.168.0.1 port 22:2: Too many authentication failures"

### DIFF
--- a/docs/tutorials/connecting-to-ev3dev-with-ssh.md
+++ b/docs/tutorials/connecting-to-ev3dev-with-ssh.md
@@ -51,7 +51,7 @@ you have configured a network connection before continuing.
 *   {: tab="Ubuntu"}
     Type the following command in a terminal window.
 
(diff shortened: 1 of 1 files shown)


tim-band:
> This seems like something that should be set in a ssh config file rather than specifying it on the command line every time.

Yes I agree, but saying "and now add this stanza to your .ssh/config" file seems way too much, saying "if you get 'too many auth failures' add this to your .ssh/config" is even more... people who know about .ssh/config will know to do it, people who don't won't want to learn it here.

Saying "if you get 'Too many authentication failures' try adding '-o PreferredAuthentication=password" seems maybe about right, but just adding it is saying even less, which is what I suggest here, but either of these seems OK to me.

